PR #5, opened 2021-07-29
Add support for no-password login.

ray-lee:
## Description

This adds support for no-password/pin logins.

1. When the `authentication` in the authentication document contains the property `inputs.password.keyboard` with the value `"No input"`, the password input is not displayed.
2. This allows the password to be `undefined` when the basic auth header is generated. The function that does this wasn't expecting this, and turned it into the string `"undefined"`.  An undefined password is now treated as `""`.
3. There were two identical functions to generate the basic auth header. These have been consolidated.
4. Added unit tests.

Screenshot:

## Motivation and Context

https://www.notion.so/lyrasis/Add-No-Pin-Password-login-support-for-CPW-3395a1065ec7479390c4a2d831a82d0b

## How Has This Been Tested?

- Checked that no-pin login works in Chrome and Firefox (using Kalama)
- Checked that login with pin continues to work (using Lyrasis)
- Added unit tests

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## CHANGELOG
 
 ### UNRELEASED CHANGES
+- Add support for no-password login.
 - Update logo/text in mobile app callouts.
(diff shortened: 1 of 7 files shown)


tdilauro (approved):
This looks good. 🚀 🎸

PR #7, opened 2021-08-03
Update Google Play Store badge to link to Palace app.

ray-lee:
## Description

Update the Google Play Store badge (in the Palace callout and the footer) to link to the Palace app instead of SimplyE.

Note: The URL was generated using the Google Play [badge generator](https://play.google.com/intl/en_us/badges/). The `pcampaignid` parameter is the default.

## Motivation and Context

https://www.notion.so/lyrasis/Update-CPW-Apps-Store-links-and-Logos-beddece52d5f44c687cb467b844d7921

## How Has This Been Tested?

Checked that clicking on the Google Play Store badge goes to the Palace app page on the Play Store (not SimplyE).

## Checklist:

### DIFF
--- a/src/components/__tests__/Footer.test.tsx
+++ b/src/components/__tests__/Footer.test.tsx
@@ -111,7 +111,7 @@ describe("toggling SimplyE Branding", () => {
     expect(googleBadge).toBeInTheDocument();
     expect(googleBadge).toHaveAttribute(
       "href",
-      "https://play.google.com/store/apps/details?id=org.nypl.simplified.simplye&pcampaignid=pcampaignidMKT-Other-global-all-co-prtnr-py-PartBadge-Mar2515-1"
+      "https://play.google.com/store/apps/details?id=org.thepalaceproject.palace&pcampaignid=pcampaignidMKT-Other-global-all-co-prtnr-py-PartBadge-Mar2515-1"
     );
 
     // my books nav link
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## CHANGELOG
 
 ### UNRELEASED CHANGES
+- Update Google Play Store badge to link to Palace app.
 - Add support for Biblioteca, Axis360, and DPLA Exchange audiobooks.
 - Add support for no-password login.
 - Update logo/text in mobile app callouts.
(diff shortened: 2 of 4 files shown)


src/components/__tests__/Footer.test.tsx
```diff
@@ -111,7 +111,7 @@ describe("toggling SimplyE Branding", () => {
     expect(googleBadge).toBeInTheDocument();
     expect(googleBadge).toHaveAttribute(
       "href",
-      "https://play.google.com/store/apps/details?id=org.nypl.simplified.simplye&pcampaignid=pcampaignidMKT-Other-global-all-co-prtnr-py-PartBadge-Mar2515-1"
+      "https://play.google.com/store/apps/details?id=org.thepalaceproject.palace&pcampaignid=pcampaignidMKT-Other-global-all-co-prtnr-py-PartBadge-Mar2515-1"
```
jonathangreen, line 114: Is the `&pcampaignid=pcampaignidMKT-Other-global-all-co-prtnr-py-PartBadge-Mar2515-1` part of the URL is doing anything? It seems like the link works without it, so maybe we just drop that part?

ray-lee: I can't find any documentation on what `pcampaignid` is used for. I can only guess that Google is using it to track usage of badges.  They don't say there's any requirement to use it, so I think it's safe to remove.

jonathangreen: Is there somewhere they ask us to use that `pcampaignid`? I was thinking maybe it was some metric NYPL had setup.

ray-lee: It's not an NYPL thing. It gets put in when you generate a URL using Google's Play Store badge generator. But they don't explain it, or say it's required.

ray-lee: The closest they get to saying anything (https://play.google.com/intl/en_us/badges/):

> Any online use of the badge must link to the Google Play store. Use the generator below to get the HTML to include in your digital marketing.

But we'd still be linking to the Play store. It doesn't say you can't modify/remove parameters.

jonathangreen: Cool! I'm good either way keeping or removing it then.

jonathangreen (approved):

PR #11, opened 2021-08-20
Implement indirect bearer token fulfillment.

ray-lee:
## Description

Implement indirect bearer token fulfillment, as described here: https://github.com/NYPL-Simplified/Simplified/wiki/OPDSForDistributors#how-it-works

- Configure indirect bearer token links to pdf and epub content to be downloadable.
- When resolving indirect links, allow receiving a download token in addition to a download URL. Previously, an indirect link was expected to resolve to a URL only.
- When downloading a book, if a network/CORS error occurs, retry the request without the `X-Requested-With` header. This non-standard (though not uncommon) header may not be whitelisted for CORS by some distributors' servers. This solves a problem downloading from BiblioBoard's S3 bucket. We can ask them to modify their CORS rules on S3, but I expect this might be a problem with other distributors as well.
- When passing headers to the function that performs HTTP requests, an `undefined` value now causes the header to be deleted. This enables the selective removal of default headers. This is used to implement the above.
- Add unit tests for the new functionality, and for some things that weren't previously tested.

## Motivation and Context

This adds support for distributors that use bearer tokens for download, e.g. Johns Hopkins, Casalini, Biblioboard, ProQuest.

Notion: https://www.notion.so/lyrasis/Add-support-for-Biblioboard-eBooks-in-CPW-13b9d749f5f24713ad82d9ccb27ac259

## How Has This Been Tested?

Verified borrowing and downloading some books from Biblioboard:

On LYRASIS:
"The Hound of the Baskervilles (Edition 1)" (epub)
"X : Volume One, Big Bad (Volume 1)" (pdf)

On Ferguson:
"Back to Basics :" (pdf)

## Checklist:

### DIFF
--- a/src/dataflow/download.ts
+++ b/src/dataflow/download.ts
@@ -10,26 +10,33 @@ export default async function downloadFile(
   type: DownloadMediaType,
   token?: string
 ) {
-  const response = await fetchWithHeaders(url, token);
+  let response;
+
+  try {
+    response = await fetchWithHeaders(url, token);
+  } catch (err) {
+    // There was a network or CORS error. In case this was a CORS error, retry the request without
+    // the X-Requested-With header. Some distributors' servers may not have this header whitelisted
+    // for CORS.
+
+    response = await fetchWithHeaders(url, token, {
+      "X-Requested-With": undefined
+    });
+  }
+
+  if (!response.ok && response.redirected) {
+    // If the response errored after a redirect, try again without the Authorization header, as it
+    // causes errors when redirected to Amazon S3.
+
+    response = await fetch(response.url);
+  }
+
   if (response.ok) {
     return downloadBlob(response, title, type);
   }
 
-  /**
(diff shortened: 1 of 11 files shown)


src/dataflow/download.ts
```diff
@@ -10,26 +10,33 @@ export default async function downloadFile(
...
-  if (response.ok) {
-    return downloadBlob(response, title, type);
+  let response;
+
+  try {
+    response = await fetchWithHeaders(url, token);
+  } catch (err) {
+    // There was a network or CORS error. In case this was a CORS error, retry the request without
+    // the X-Requested-With header. Some distributors' servers may not have this header whitelisted
+    // for CORS.
```
tdilauro, line 20: Any thoughts on how to surface this so that we know to ask a distributor to fix?

tdilauro (approved):
Looks good to me.

PR #12, opened 2021-08-23
Create sync.yml

jonathangreen:
## Description

Add github action that syncs a branch (`nypl/dev`) on this repo with NYPL every night. This makes it easier to pull in changes from upstream and resolve any merge conflicts, when we want to pull in changes. 

We are doing a similar sync on:
- circulation 
- circulation-core 
- circulation-admin 
- library-registry
- library-registry-admin

I thought it probably makes sense to do this here as well.

### DIFF
--- a/.github/workflows/sync.yml
+++ b/.github/workflows/sync.yml
@@ -1,0 +1,39 @@
+name: Sync branch with NYPL
+on:
+  schedule:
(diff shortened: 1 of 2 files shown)


tdilauro (approved):
🥁

PR #14, opened 2021-09-01
Revert branding changes.

ray-lee:
## Description

Reverts changes that renamed SimplyE to Palace, and linked to the Palace apps in the app stores.

## Motivation and Context

This allows other changes on main to be deployed now, before the Palace apps are ready to be publicized. 

## How Has This Been Tested?

Verified that the mobile app callout on the book detail page and in the footer now refer to SimplyE instead of Palace, and the app store badges link to the SimplyE app instead of Palace.

## Checklist:


src/components/bookDetails/index.tsx
```diff
@@ -13,7 +13,7 @@ import ReportProblem from "./ReportProblem";
 import Head from "next/head";
 import { H1, H2, H3, Text } from "components/Text";
 import MediumIndicator from "components/MediumIndicator";
-import PalaceLogo from "components/PalaceLogo";
+import SimplyELogo from "components/SimplyELogo";
```
tdilauro, line 16: Assuming that we still have this component.

tdilauro (approved):
Looks good. 📚

src/components/bookDetails/index.tsx
ray-lee: Yup, I never deleted it.

PR #15, opened 2021-09-13
Upgrade next to 10.0.8.

ray-lee:
## Description

Upgrade next from 10.0.5 to 10.0.8.

## Motivation and Context

This should fix the `Cannot find module 'sharp'` error on installation: https://github.com/vercel/next.js/issues/20456

## How Has This Been Tested?

I verified that the app still installs and runs. I never saw the error on my dev environment, so can't verify that it's actually fixed, but 10.0.8 does include the claimed fix.

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## CHANGELOG
 
 ### UNRELEASED CHANGES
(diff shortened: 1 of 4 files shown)


jonathangreen (approved):
Looks good! If you want to merge I'll deploy the new container out to the test server.

PR #31, opened 2022-10-13
Fix for redirect URLs incorrectly being identified as full URL

jonathangreen:
## Description

From this notion ticket:
https://www.notion.so/lyrasis/Correct-Columbia-CPW-configuration-to-match-the-QA-config-Borrow-Only-76e5c0578a0943f2bebbb07ac11909f6

In Columbia CPW instance users were not getting redirected properly when borrowing.

## Motivation and Context

When borrowing books, users aren't getting redirected after logging in. This is happening because of a new check added in the code, the does not redirect when a URL contains `http`, however some URLs in the Columbia instance have books with `http` in the identifier. So they were being detected as full URLs. 

For example a book with this identifier: 
`https%3A%2F%2Fcolumbia.edu.thepalaceproject.org%2F190150%2Fworks%2FURI%2Furn%3Ax-internet-archive%3Aebooks-app%3Aitem%3Aldpd_12584146_000`

## How Has This Been Tested?

- Tested locally, by pointing to Columbia CM 
- Added some tests for this case

### DIFF
--- a/src/auth/__tests__/LoginWrapper.test.tsx
+++ b/src/auth/__tests__/LoginWrapper.test.tsx
@@ -83,4 +83,46 @@ describe("redirects when user becomes authenticated", () => {
       shallow: true
     });
   });
+
+  test("uses homepage when nextUrl start with http", () => {
+    render(<LoginWrapper />, {
+      user: {
+        isAuthenticated: true
+      },
+      router: {
+        query: {
+          nextUrl: "http://test.com/test123"
+        }
+      }
+    });
+
+    expect(mockPush).toHaveBeenCalledTimes(1);
+    // redirects to home page instead of nextUrl
+    expect(mockPush).toHaveBeenCalledWith("/testlib", undefined, {
+      shallow: true
+    });
+  });
+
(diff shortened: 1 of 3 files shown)


src/auth/__tests__/LoginWrapper.test.tsx
```diff
@@ -83,4 +83,46 @@ describe("redirects when user becomes authenticated", () => {
...
+    });
+
+    expect(mockPush).toHaveBeenCalledTimes(1);
+    // redirects to home page instead of nextUrl
+    expect(mockPush).toHaveBeenCalledWith("/testlib", undefined, {
+      shallow: true
+    });
+  });
+
+  test("uses nextUrl when url contains http", () => {
```
tdilauro, line 106: Minor: Might be slightly confusing. Maybe "uses nextUrl when url contains, but does not start with, http".

tdilauro (approved):
LGTM! 🎸 

Just one minor suggestion, but not a show-stopper.

ray-lee (approved):
Looks good.

PR #42, opened 2024-12-02
Working/upgrade without legacy peer deps

PoorBillyPilgrim:
## Description

Major updates and/or replacements to several key packages. Notably, the project now uses React 18 and Next.js 15. All older tests using enzyme have been removed or replaced with jest. The dependency tree was also cleaned up which no longer required `--legacy-peer-deps` when installing.

## Motivation and Context

These updates removed any identifiable security vulnerabilities and any new functionality can be built without the worry of using older or deprecated libraries.

## How Has This Been Tested?

Tests were updated to account for package updates or replacements.

## Checklist:


tdilauro (approved):
This looks good! 🚀 

I was able to build and run from docker container, so I'm going to approve this one and merge it.

PR #44, opened 2025-01-13
Update default libraries configuration.

tdilauro:
## Description

Updates the repo's default `libraries` configuration to use a single Palace demo library. An alternative configuration that loads all non-hidden libraries from the Palace Library Registry is present, but commented out, to allow for easy switching when the need arises.

## Motivation and Context

- The previous default was to load `libraries` from the SimplyE, rather than Palace, registry. This led to loading many libraries that appear not to be active any longer, resulting in lots of error messages.
- Loading only a single library by default should speed up application start up. The configuration can be temporarily updated locally when the need arises.

## How Has This Been Tested?

- Functionality manually tested locally.
- All tests pass locally.

## Checklist:

- [X] All new and existing tests passed.


PoorBillyPilgrim (approved):
👍

PR #46, opened 2025-02-11
PP-954 : add distributor name to book details page

PoorBillyPilgrim:
## Description

Added the name of the distributor (e.g. OverDrive, Palace Marketplace, etc.) to the book details page.
## Motivation and Context

This change matches those made to the iOS app that already display the distributor.

## How Has This Been Tested?

The `providerName` is only rendered on the details page and mimics how publisher and published are parsed and rendered. Its addition only affects the details page. By adding `getProviderName` to `entryToBook`, testing revealed several tests that pipe test entries through this function, some of which were missing the `unparsed` property (a property which should always be provided by opds-feed-parser). Changes were made to `getProviderName` to account for this. So, a missing unparsed or distributor value won't throw an error.

## Checklist:


tdilauro (approved):
This looks good! 🚀 

Should we add a ticket for parsing the BibFrame data?

PoorBillyPilgrim:
@tdilauro Not sure it's necessary. From just looking at the iOS code (assuming this is an up-to-date representation of a book details page in the Palace ecosystem), distributor is the only bibframe data we're pulling. 

That said, it's easy enough to parse the xml for any bibframe data. It wouldn't be a big lift if we just to have it around in case.

PoorBillyPilgrim:
@tdilauro you know, I gave it some thought. I'm just gonna wrap up that bibframe data in this ticket.

PoorBillyPilgrim:
@tdilauro I added a Map for any bibframe tags. It's not a huge addition, but I do think it's an improvement. At the least, having the xml2js types provides some better clarity.

src/test-utils/fixtures/raw-opds-entry.ts
tdilauro, line 1: Not a showstopper, but do we need to add the `<bibframe:distribution bibframe:PublisherName="Overdrive"/>` here?

PoorBillyPilgrim: No, not all all. Unintended addition -- thanks for catching.

PR #48, opened 2025-03-18
Allows configuration of media support default for unmatched content types (PP-2190)

tdilauro:
## Description

Allows configuration of a default media support value to be used when the `content type` or `indirect link type`/`content type` pair do not match an entry in the configuration. This is set in the `media_support.default` property in the configuration file. For example:

```
...
media_support:
  default: show

  application/audiobook+json: redirect

  application/vnd.readium.lcp.license.v1.0+json:
    application/epub+zip: redirect
...
```

The behavior will be as follows:
- If the the `content type` or `indirect link type`/`content type` pair match, then the value specified for the `content type` will be used.
- Else, if a `default` is specified, that default value will be used.
- Else, "unsupported" will be used.

## Motivation and Context

As we experiment with different configurations, it will be easier to reconfigure defaults, if we can do it through configuration, rather than code, changes.

[Jira [PP-2190](https://ebce-lyrasis.atlassian.net/browse/PP-2190)]

## How Has This Been Tested?

- Added new test cases.
- All tests pass locally.
- Review CI tests.

## Checklist:

[PP-2190]: https://ebce-lyrasis.atlassian.net/browse/PP-2190?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### UNRELEASED CHANGES
 
+- Allows configuration of media support default for unmatched content types. (PP-2190)
 - distributor added to book details page
 - Log auth document URL when Catalog Root link missing. (PP-1209)
 - Update default libraries config.
--- a/src/interfaces.ts
+++ b/src/interfaces.ts
@@ -31,7 +31,12 @@ export type OpenEbooksConfig = {
   defaultLibrary: string;
 };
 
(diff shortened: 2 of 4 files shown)


PoorBillyPilgrim (approved):
👍

PR #51, opened 2025-03-24
CI improvements. (PP-1222)

tdilauro:
## Description

This update does not change any application code.

- Build containers after successful tests.
- Add branch, commit, PR merge, and release container builds/tags.
- Deprecate change log.
- Adds a GitHub PR template.
- Adds a GitHub release configuration.

## Motivation and Context

- Move CI toward workflows more in line with our other repositories.
- Better availability of docker container images for testing and deployment.

[Jira PP-1222]

## How Has This Been Tested?

- Manual testing in a separate repository.
- Will validate changes using the pushes, PR, and release associated with this PR.


.github/workflows/test-and-build.yml
```diff
@@ -0,0 +1,194 @@
...
+
+      - name: Build and push Docker image
+        uses: docker/build-push-action@v6
+        with:
+          context: .
+          file: ./Dockerfile
+          push: ${{ env.IMAGE_PUSH_ENABLED }}
+          tags: ${{ steps.docker-image-tags.outputs.tags }}
+          cache-from: type=local,src=/tmp/.buildx-cache
+          cache-to: type=local,dest=/tmp/.buildx-cache
```
jonathangreen, line 194: Minor: This will work, but I believe the cache will grow unbounded, until its 10Gb and then github will automatically delete it, then the process will start again. You might instead want to consider using a [registry cache](https://docs.docker.com/build/ci/github-actions/cache/#registry-cache) or the [github cache](https://docs.docker.com/build/ci/github-actions/cache/#github-cache), since they both provide better targeted cache management.

This work could be a follow up though, it becomes a problem. Just wanted to note those other cache sources, so you knew about them.

jonathangreen (approved):
Looks good, nice improvements to CI here 🥇

PR #53, opened 2025-04-02
Added format to book detail page. (PP-2191)

PoorBillyPilgrim:
## Description

Added book format (ePub, PDF, or Audiobook) to detail pages

## Motivation and Context

This change matches the mobile apps which already displayed format.

[Jira PP-2191](https://ebce-lyrasis.atlassian.net/jira/software/c/projects/BBL/boards/14?selectedIssue=PP-2191)

## How Has This Been Tested?

Added two tests to infer book format

## Checklist:


PoorBillyPilgrim:
@jonathangreen pinging you on this PR cause Tim is shown as vacationing on slack. If you're unable to get around to this, no biggie.

src/types/opds1.ts
```diff
@@ -94,11 +96,29 @@ export const AccessRestrictionAudiobookMediaType =
...
+
 export type ReadOnlineMediaType =
   | typeof ExternalReaderMediaType
   | typeof AxisNowWebpubMediaType
   | typeof OverdriveEbookMediaType;
 
+export type EpubFormatType =
+  | typeof EpubMediaType
+  | typeof KepubMediaType
+  | typeof AxisNowWebpubMediaType;
```
tdilauro, line 113: This works, but if you want `EpubFormatType` to track changes to `EPUB_MEDIA_TYPES` over time without manual updating, you can you something like this: 

```
  export type EpubFormatType = typeof EPUB_MEDIA_TYPES[number];
```

src/dataflow/opds1/parse.ts
```diff
@@ -101,6 +107,50 @@ function parseFormat(format: OPDSAcquisitionLink | OPDSIndirectAcquisition) {
...
+
+      if (mimeType === PdfMediaType) {
+        return "PDF";
+      }
+    }
+
+    return undefined;
+  } else {
+    return undefined;
+  }
```
tdilauro, line 151: Since we're returning on match I think we can drop the `else` and the `return undefined` inside the `if` and just have a final `return undefined;` at the end of the function.

tdilauro (approved):
Code here looks good! 🎸

A couple small suggestions in my code comments below and a request related to the titles/content of our PRs here in this comment.

We should use meaningful names on our PRs. If one of our future selves needs to search or scan through old PRs, "Pp 2191" won't be very helpful. And for people looking at this as open source (or us in a future after Jira), our Jira tickets might be meaningless.

Also, with #51 (CI Improvements), we introduced changes that will allow us to cut releases a little more easily. The notes we autogenerate for those releases will include the PR title, so having good titles will be helpful there, too. For this one, something like the following, would be more helpful:
- `Added format to book detail page. (PP-2191)`

IN general, I think that our PR titles and description should have enough information in them to convey the gist of what changed and why, even in the absence of access to the issue tracker.

CHANGELOG.md
```diff
@@ -2,8 +2,9 @@
 
 ### UNRELEASED CHANGES
 
+- Add book format (ePub, PDF, Audiobook) to book details page (PP-2191)
 - Change test/build GitHub workflow to build after tests. (PP-1222)
-  - Once this change is merged, we will abandon this change log and use GitHub release notes instead.
+- Once this change is merged, we will abandon this change log and use GitHub release notes instead.
```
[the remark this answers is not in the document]
PoorBillyPilgrim: I was confused by the presence of the "Change test/build..." comment added after. I'll stop updating the changelog moving forward.

PR #54, opened 2025-04-04
Render modals in center of screen by default (PP-2243)

PoorBillyPilgrim:
## Description

Updated custom Modal component to render in the middle of the screen. Before, the modal was rendering at the bottom of the page. This might have been an unintended change after updating to ariakit from reakit.

## Motivation and Context

Separating the modal so far from the event origin was bad UX.

[Jira PP-2243](https://ebce-lyrasis.atlassian.net/browse/PP-2243)

## How Has This Been Tested?

Updated the sign out modal test to expect the new styling changes whenever the modal is visible.

## Checklist:

### DIFF
--- a/src/components/__tests__/SignOut.test.tsx
+++ b/src/components/__tests__/SignOut.test.tsx
@@ -20,8 +20,16 @@ test("Shows modal on click", async () => {
   await user.click(signOut);
 
   const modal = await screen.findByLabelText("Sign Out");
+  screen.debug(modal);
 
-  expect(modal).toHaveStyle("display: block");
+  expect(modal).toHaveStyle({
+    display: "block",
+    position: "fixed",
+    height: "fit-content",
+    maxWidth: "400px",
+    inset: "0.75rem",
+    margin: "auto"
+  });
   expect(modal).toBeVisible();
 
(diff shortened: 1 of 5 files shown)


src/components/__tests__/SignOut.test.tsx
```diff
@@ -20,8 +20,16 @@ test("Shows modal on click", async () => {
   await user.click(signOut);
 
   const modal = await screen.findByLabelText("Sign Out");
+  screen.debug(modal);
```
tdilauro, line 23: Some leftover debugging here.

src/components/bookDetails/ReportProblem.tsx
```diff
@@ -55,7 +55,7 @@ const ReportProblem: React.FC<{ book: AnyBook }> = ({ book }) => {
         dialog={dialog}
         label="Report a problem"
         hide={cancel}
-        sx={{ width: "100%", maxWidth: 600 }}
+        sx={{ maxWidth: 600 }}
```
tdilauro, line 58: Minor: Might as well add units here, since we're touching it.

tdilauro (approved):
LGTM! 🥇 

A couple small things below.

PR #55, opened 2025-04-05
Grab book format from fulfillment link, if available (PP-2191)

PoorBillyPilgrim:
## Description

Grab book format from a book's list of fulfillment links, if list exists

## Motivation and Context

While working on a separate ticket, I noticed that books that had been fulfilled would have their indirect acquisition links removed, and so the book format would be removed from the book detail page.

[Jira PP-2191]

## Checklist:

### DIFF
--- a/src/dataflow/opds1/parse.ts
+++ b/src/dataflow/opds1/parse.ts
@@ -124,25 +124,33 @@ const linearizeAcquisitions = ia =>
  * If an entry has "application/epub+zip" followed by "application/pdf",
  * epub takes precedence. So, get first mimetype.
  * @see https://github.com/io7m/opds-acquisition-spec?tab=readme-ov-file#linearized-acquisitions
+ *
+ * If a book has been fulfilled in some way, then the mimetype can
+ * be derived from there
  */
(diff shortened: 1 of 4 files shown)


src/dataflow/opds1/parse.ts
```diff
@@ -124,25 +124,33 @@ const linearizeAcquisitions = ia =>
  * If an entry has "application/epub+zip" followed by "application/pdf",
  * epub takes precedence. So, get first mimetype.
  * @see https://github.com/io7m/opds-acquisition-spec?tab=readme-ov-file#linearized-acquisitions
+ *
+ * If a book has been fulfilled in some way, then the mimetype can
+ * be derived from there
```
tdilauro, line 129: Minor: "... from there"

Wnat does "there" mean here (sorry for that confusing sentence)?

tdilauro (dismissed):
Looks good! 🚀

tdilauro (commented):
I think we should have a test that distinguishes these cases.

tdilauro (approved):
Looks good! 🏁

PR #56, opened 2025-04-06
Add return button to My Books page for fulfillable books (PP-2283)

PoorBillyPilgrim:
## Description

When a user navigates to their "My Books" page, they will now see an option to return their book from that list if applicable. 

## Motivation and Context

This change matches the actions a user can take from the book details page, allowing users to return a book without having to drill into the book details page itself.

[Jira PP-2283](https://ebce-lyrasis.atlassian.net/jira/software/c/projects/BBL/boards/14?selectedIssue=PP-2283)

## How Has This Been Tested?

I added a test to expect the return button if a book is fulfillable

## Checklist:

### DIFF
--- a/src/components/BookList.tsx
+++ b/src/components/BookList.tsx
@@ -243,19 +243,26 @@ const BookListCTA: React.FC<{ book: AnyBook }> = ({ book }) => {
     const singleFulfillment =
       showableFulfillments.length === 1 ? showableFulfillments[0] : undefined;
 
-    if (singleFulfillment && !shouldRedirectUser) {
-      return (
-        <FulfillmentButton
-          details={singleFulfillment}
-          book={book}
-          isPrimaryAction
-        />
-      );
-    }
     return (
-      <NavButton variant="link" bookUrl={book.url} iconRight={ArrowForward}>
-        View Book Details
-      </NavButton>
+      <>
+        <CancelOrReturn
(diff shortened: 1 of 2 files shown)


src/components/BookList.tsx
```diff
@@ -243,19 +243,26 @@ const BookListCTA: React.FC<{ book: AnyBook }> = ({ book }) => {
...
-      <NavButton variant="link" bookUrl={book.url} iconRight={ArrowForward}>
-        View Book Details
-      </NavButton>
+      <>
+        <CancelOrReturn
+          url={book.revokeUrl}
+          loadingText="Returning..."
+          id={book.id}
+          text="Return"
+        />
```
tdilauro, line 253: Do we need a logged in user (or to be on an auth-protected route for this behavior? Does this need to be protected by authentication? Could we show a return button to a user who is not currently logged in? Or would we just never do that (e.g., because no books with revoke URLs will be in the regular catalog feed)?

tdilauro (approved):
This looks good. One important question below, but already approved, in case it is not an issue.

src/components/BookList.tsx
PoorBillyPilgrim: @tdilauro FWIW, this is logic borrowed from the book details page, so we do render a return button on route that is not auth protected. This change only adds an additional place in which a return can be made.

PR #64, opened 2025-04-30
Remove support for unavailable AxisNow decryptor. (PP-941)

tdilauro:
## Description

Removes support for unavailable AxisNow decryptor.

## Motivation and Context

The `@nypl-simplified-packages/axisnow-access-control-web` dependency is proprietary and is not available to this project, so the code to support it is superfluous and adds complexity.

[Jira PP-941]

## How Has This Been Tested?

- Manually tested in local environment.
- All tests pass locally.
- [CI tests](https://github.com/ThePalaceProject/web-patron/actions/runs/14756603316) pass.

## Checklist:

- [X] All new and existing tests passed.


src/components/WebpubViewer.tsx
PoorBillyPilgrim, line 1: @tdilauro I'm seeing this component only used by the `pages/[library]/read/[bookUrl].tsx`, which was deleted. Are we keeping this around in case we want to reuse the CSS for a future reader in the app?

tdilauro: Okay. All set. 🔥

PoorBillyPilgrim (approved):

PR #65, opened 2025-05-05
PP-2463 : [bugfix] pressing enter from login barcode or password input would not initiate login

PoorBillyPilgrim:
## Description

The login screen has a button that toggles the visibility for a user's password. The type for this button has been explicitly set as `button`.

## Motivation and Context

Any buttons nested within a `<form/>` will have `type="submit"` by default if no type is specified. The original click handler for the password visibility toggle overrode the form's default behavior, so any submit event within the form would not work as intended and would only toggle password visibility. Adding `type="button"` to the password visibility button separated the button's behavior from the form and allows an `Enter` keydown to submit from anywhere within the form unless a user is focused on the password visibility button.

[Jira PP-2463](https://ebce-lyrasis.atlassian.net/browse/PP-2463)

## Checklist:

- [N/A] I have updated the documentation accordingly.


tdilauro (approved):
Looks good! 🚀

PR #114, opened 2026-04-06
Add SAML logout support (PP-3453)

tdilauro:
## Description

SAML logout support with a supporting refactor:

- **SAML logout flow** — Now detects a SAML logout link on the auth method, calls the server-side logout endpoint, then navigates to `/signed-out`. Falls back to the existing redirect-to-sign-out-page-based flow if no logout link is present.
- **`performLogoutRequest` helper** — extracted from the OIDC logout path so both OIDC and SAML share the same endpoint-call-and-navigate logic.

## Motivation and Context

Enables this application to interact with the SAML logout capabilities in the Palace Manager,
following the same pattern as the OpenID Connect logout support introduced in #106.

[Jira PP-3453]

## How Has This Been Tested?

- Manual testing in a local development environment.
- New and updated tests for this new support.
- All tests pass locally.
- [CI checks](https://github.com/ThePalaceProject/web-patron/actions/runs/24036393446) pass.

## Checklist:

### DIFF
--- a/src/components/SignOut.tsx
+++ b/src/components/SignOut.tsx
@@ -5,8 +5,8 @@ import Button from "./Button";
 import Stack from "./Stack";
 import useUser from "components/context/UserContext";
 import { styleProps } from "./Button/styles";
-import { OPDS1, ClientOidcMethod } from "interfaces";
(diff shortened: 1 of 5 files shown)


PoorBillyPilgrim (approved):
👍

PR #118, opened 2026-04-14
Dynamic runtime registry fetch support (PP-3114)

tdilauro:
## Description

- Changes library setup for libraries configured via `libraries: <url-string>` (deprecated) and `registries[].url` to occur at runtime (e.g., `npm run start`), rather than at build time.
- Adds support and optimizations for crawlable paginated library registry feeds, when configured and when required `order=modified` sort facet is advertised.
- Creates hooks for computing and validating slugs.

## Motivation and Context

Improves service start up performance by instantiating library pages during runtime, rather than build time. Our deployment currently builds the app in a docker container, which could not complete startup until the build was finished.

Improves runtime performance by incrementally loading updated registry libraries during refresh, reducing the load on each instance and on the configured registries.

Lays groundwork for future registry improvements.

[Jira PP-3114]

## How Has This Been Tested?

- Checks (tests, type checking, linter) pass locally.
- CI tests pass.

## Checklist:

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,7 +1,7 @@
 module.exports = {
   overrides: [
     {
-      files: ["**/__tests__/**/*.{ts,tsx}", "src/test-utils/**/*.{ts,tsx}"],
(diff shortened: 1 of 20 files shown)


PoorBillyPilgrim (approved):
Looks great @tdilauro -- good stuff 🚀

PR #119, opened 2026-04-16
Statically configured libraries load at runtime, not build time (PP-4083)

tdilauro:
## Description

- Moves static library loading from build time to runtime. Libraries defined in the `libraries` section of `CONFIG_FILE` are now read and cached on first server request rather than baked into the build artifact.
- Registry-sourced libraries continue to be fetched and cached per the existing registry refresh logic.

## Motivation and Context

- Waiting to generate libraries pages for the statically-specified libraries improves build performance, allowing the container to become ready more quickly.
- Library configuration (auth doc URLs, titles) was previously embedded in the build artifact, requiring a rebuild to reflect config changes. This change allows the server to serve the current library list from `CONFIG_FILE` without a rebuild.

[Jira PP-4083]

## How Has This Been Tested?

- New and updated tests.
- All checks (tests, type checker, linter) pass locally.
- Build succeeds locally.
- [CI checks](https://github.com/ThePalaceProject/web-patron/actions/runs/24487198022) pass.

## Checklist:

### DIFF
--- a/msw/handlers/index.ts
+++ b/msw/handlers/index.ts
@@ -1,11 +1,5 @@
 import { http, HttpResponse } from "msw";
-import { APP_CONFIG } from "utils/env";
 import { PATRON_PROFILE_FIELDS } from "types/patronProfile";
-
-export const openEBackendUrl = APP_CONFIG.libraries["app"]?.authDocUrl.replace(
-  "/groups",
-  ""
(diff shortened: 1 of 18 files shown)


jonathangreen (approved):
🚀 This is a big improvement

PR #121, opened 2026-04-16
Deployed container startup time improvements (PP-4001)

tdilauro:
## Description

This PR introduces three changes to reduce container startup time:
- **Enables SWC for production builds.** Deletes `babel.config.js`, whose presence forced Next.js to use Babel for compilation. The Babel preset (needed by Jest for Theme UI's emotion JSX transform) is moved inline into `jest.config.js` and `jest.config.node.js` so the test suite is unaffected.
- **Adds `SKIP_BUILD_TIME_CHECKS` flag.** A new `SKIP_BUILD_TIME_CHECKS` env var (`false` by default in the Dockerfile) allows deployments to skip ESLint and TypeScript checks during `next build`. When enabled via the deployment config, this eliminates the lint/typecheck phase from container startup entirely.
- **Removes deprecated `next-transpile-modules`.** Replaces the `next-transpile-modules` wrapper with Next.js's built-in `transpilePackages` option (available since Next.js 13). This was needed for the switch from Babel to SWC and has the added benefit of removing a dependency.

## Motivation and Context

Container startup was slow (usually taking over 3.5 minutes) due to long app build times (we currently need to build at container start to some parts of the active configuration). The build used Babel compilation and performed linting and type checking — all of which added latency on every deploy / restart. 
- SWC is significantly faster than Babel for production builds.
- Our CI already performs linting and type checking, so wasting ~50 seconds on it for every startup should be avoided.

[Jira PP-4001]

## How Has This Been Tested?

- Manual testing with `npm run build` and `npm run start` in my local dev environment.
- All checks pass locally.
- CI tests pass.

## Checklist:


PoorBillyPilgrim (approved):
Looks good 🚀

PR #122, opened 2026-04-17
Fix registry cache race condition (PP-4102)

tdilauro:
## Description

Fixes a registry cache race condition by tracking the in-progress crawl instead of returning an empty cache immediately. Duplicate overlapping calls now await the in-progress call.

## Motivation and Context

After `npm run build && npm run start` (I did not see this bug with `npm run dev`), the first navigation to a registry-sourced library page intermittently returned a 404 ("No authentication document url is configured"). A subsequent visit would succeed.

ISR was firing two near-simultaneous `getStaticProps` calls on a cold page load. The first call set `lastAttemptedFetch` synchronously before its async crawl completed. The second call hit the `shouldRefresh` min-interval guard, returned immediately with an empty cache, and caused the 404. The error page would set revalidation, so subsequent visits worked once the first crawl had finished.

The fix tracks in-progress crawl promises, so subsequent requests await the active crawl rather than bailing early.

[Jira PP-4102]

## How Has This Been Tested?

- Manual local testing with both `npm run build-and-start` and `npm run dev`.
- New tests (happy path, failure path) specifically for that condition. Happy path test failed against the pre-fix code.
- All checks pass locally.
- [CI checks](https://github.com/ThePalaceProject/web-patron/actions/runs/24567733387) pass.

## Checklist:

### DIFF
--- a/src/server/__tests__/libraryRegistry.test.ts
+++ b/src/server/__tests__/libraryRegistry.test.ts
@@ -949,6 +949,35 @@ describe("concurrent first-fetch coalescing", () => {
     // Only one network request despite two concurrent callers.
     expect(fetchMock).toHaveBeenCalledTimes(1);
   });
+
+  it("concurrent callers both get an empty result when the in-progress crawl fails", async () => {
+    let resolveFetch!: () => void;
+    const fetchGate = new Promise<void>(resolve => {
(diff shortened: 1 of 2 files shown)


PoorBillyPilgrim (approved):
🚀

PR #124, opened 2026-04-23
Add runtime config module with ArkType validation (PP-4124)

tdilauro:
## Description

**Note:** This is the first in a series of PRs to move fully to runtime configuration. Other associated PRs will stack on this one.

Adds the server-side runtime config module and supporting pieces that subsequent PRs will wire into the app. No existing code is changed; nothing imports these new files yet.

Specifically, this PR adds:
- A server-side config loader that reads and parses `CONFIG_FILE` at runtime on first request and caches the result for the process lifetime. `ArkType` validates the raw YAML structure upfront so malformed configs produce a clear error rather than a confusing downstream failure. Per-slug library-entry errors are handled separately so their messages can include the offending slug in the path.
- A React context and `useAppConfig()` hook for components to consume config values, and a safe fallback config for use before `pageProps` is available (e.g. during SSR).
- Support for the `staticLibraries` "shape" in `AppConfig` — object-format `libraries` entries mapping slugs directly to auth document URLs.
- A config file reference in the README (options table and media support explanation), so the deprecation warning for the legacy string `libraries` format has somewhere concrete to point.

## Motivation and Context

The long-term goal is to read configuration at server startup rather than baking its contents into the build artifact, so that config changes don't require a full rebuild and redeploy.

This PR adds the new module without activating it; a follow-on PR will switch consumers over.

[JIRA PP-4124]

## How Has This Been Tested?

- Includes unit tests for the new functionality.
- All check pass locally.
- [CI checks](https://github.com/ThePalaceProject/web-patron/actions/runs/24856257601) pass.

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,8 @@ __To have your library added to the demo, register it with NYPL's Library Regist
   <!-- * [Demo](#community-demo) -->
 - [Configuring the App](#configuring-the-app)
   - [Configuration File](#configuration-file)
+    - [Configuration Options](#configuration-options)
+    - [Media Support](#media-support)
   - [Environment Variables](#environment-variables)
   - [Manager, Registry, and Application Configurations](#manager--registry--and-application-configurations)
   - [Libraries and Registries Configuration Settings](#libraries-and-registries-configuration-settings)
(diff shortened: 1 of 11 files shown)


src/server/appConfig.ts
```diff
@@ -0,0 +1,257 @@
...
+const RawConfigSchema = type({
+  // Intentionally permissive: any non-string value silently defaults.
+  "instance_name?": "unknown",
+  "companion_app?": "string",
+  "show_medium?": "boolean",
+  "bugsnag_api_key?": "string",
+  "gtmId?": "string",
+  "registries?": RegistryEntrySchema.array(),
+  "libraries?": "string | Record<string, unknown>",
+  "media_support?": "Record<string, string>",
```
jonathangreen, line 43: The [existing config](https://github.com/ThePalaceProject/web-patron/blob/main/community-config.yml) structure allows nested types for indirect acquisitions: 

https://github.com/ThePalaceProject/web-patron/blob/902d6d59268332886fb1adbd0b5dd7c9cc617558/community-config.yml#L43-L46

I think the type here needs to take that into account.

src/server/appConfig.ts
```diff
@@ -0,0 +1,257 @@
...
+  "refreshMaxInterval?": "number"
+});
+
+const RawConfigSchema = type({
+  // Intentionally permissive: any non-string value silently defaults.
+  "instance_name?": "unknown",
+  "companion_app?": "string",
+  "show_medium?": "boolean",
+  "bugsnag_api_key?": "string",
+  "gtmId?": "string",
```
jonathangreen, line 40: Every other key here is `snake_case`, it would be nice to align `gtmId` if we can. It looks like it was previously treated as `snake_case` in the example config:

https://github.com/ThePalaceProject/web-patron/blob/902d6d59268332886fb1adbd0b5dd7c9cc617558/community-config.yml#L55-L56

README.md
```diff
@@ -65,9 +67,30 @@ __To have your library added to the demo, register it with NYPL's Library Regist
...
+
+| Key | Type | Default | Description |
+|-----|------|---------|-------------|
+| `instance_name` | string | `"Patron Web Catalog"` | Name used in error tracking and debug output. Not patron-facing. |
+| `companion_app` | `"simplye"` \| `"openebooks"` | `"simplye"` | Selects which companion mobile app to reference in redirect prompts. |
+| `show_medium` | boolean | `true` | Whether to display the medium (e-book, audiobook, etc.) label on book cards. |
+| `bugsnag_api_key` | string | — | Bugsnag project API key. Omit to disable error tracking. |
+| `gtmId` | string | — | Google Tag Manager container ID (e.g. `GTM-XXXX`). Omit to disable analytics. |
+| `media_support` | mapping | `{}` | Per-MIME-type rendering mode. See [Media Support](#media-support) below. |
+| `libraries` | mapping or string | — | Static library definitions, or a deprecated registry URL string. See [Libraries and Registries Configuration Settings](#libraries-and-registries-configuration-settings). |
```
jonathangreen, line 82: Might be worth noting here that the string form is deprecated.

jonathangreen (approved):
Looks good! I added couple questions / comments to consider before merging.

src/server/appConfig.ts
jonathangreen: I reviewed this PR first, but it looks like the follow up PR addresses this:
https://github.com/ThePalaceProject/web-patron/pull/125/changes#diff-5e10033ed4c08cfb900c998fd81feeb91338a52fc01f227d6a790bb919e617b7R44

PR #125, opened 2026-04-23
Wire up new runtime config module (PP-4124)

tdilauro:
**Note:** This PR is stacked on #124 and should not be merged until that one goes in.

## Description

Activates the runtime config system introduced in the companion PR (`chore/runtime-config-module`) and removes the build-time config machinery it replaces.

**Wiring:**
- All page and API routes now receive `AppConfig` via `getAppConfig()` (reads and parses `CONFIG_FILE` at runtime, cached in memory) rather than from a build-time baked environment variable
- `_app.tsx` wraps the component tree with `AppConfigContext.Provider`, initialises Bugsnag, and applies media support settings from the runtime config
- `getLibraries` in `libraryRegistry.ts` now reads static libraries directly from `AppConfig.staticLibraries` rather than re-parsing `CONFIG_FILE` independently

**New config property:**
- `staticLibraries` is a first-class config key for inline static library definitions, replacing the object form of `libraries`
- Using `libraries` as an object now emits a deprecation warning and is rejected if `staticLibraries` is also present
- `community-config.yml` and README updated to reflect the new property and mark both deprecated `libraries` forms

**Build / Docker:**
- Bugsnag webpack plugin removed; Bugsnag is now initialised at runtime from `AppConfig`
- Dockerfile updated to multi-stage build with `output: "standalone"`, reducing image size

**Dead code removed:**
- `src/config/fetch-config.js` and its type declaration and tests
- `src/server/staticLibraries.ts` and its tests (superseded by `AppConfig.staticLibraries`)
- `SKIP_BUILD_TIME_CHECKS` env var reference removed from README

## Motivation and Context

Config was previously baked into the build artifact at `next build` time, requiring a full rebuild to change any setting. This change moves config resolution to runtime so that a container image can be reconfigured (or repointed at a different `CONFIG_FILE`) without rebuilding.

[Jira PP-4124]

## How Has This Been Tested?

- All checks pass locally.
- [CI checks](https://github.com/ThePalaceProject/web-patron/actions/runs/24861760146) pass.

## Checklist:


src/pages/index.tsx
```diff
@@ -1,14 +1,42 @@
...
+  return <MultiLibraryHome />;
+};
+
+export const getServerSideProps: GetServerSideProps<HomeProps> = async ctx => {
+  let appConfig: AppConfig;
+  try {
+    appConfig = await getAppConfig();
+  } catch (e) {
+    if (e instanceof AppSetupError) {
+      return { props: { appConfig: FALLBACK_APP_CONFIG } };
```
jonathangreen, line 26: Instead of falling back here, do we want to surface an error? That way it would be clear an error is occurring. If we served an error 500 or something instead it would be clear to load balancer or smoke tests that something has gone sideways.

jonathangreen (approved):
Looks good. One comment to consider before merging.

src/pages/index.tsx
jonathangreen: I guess I needed to look at all these, it looks like this was handled in https://github.com/ThePalaceProject/web-patron/pull/126, so you can ignore this.

PR #126, opened 2026-04-24
Trigger configuration load on startup (PP-4124)

tdilauro:
**Note:** This PR is stacked on #125, which is, in turn, stacked on #124, and should not be merged until those have landed.

## Description

Trigger configuration load success or failure immediately upon startup:
- Uses the Instrumentation API to load and validate the app configuration at server startup, rather than on the first request.
- If config loading fails — missing `CONFIG_FILE`, unreachable URL, invalid YAML — the error message is printed to stderr and the process fails fast, exiting with an error before serving any requests.

## Motivation and Context

This change makes misconfiguration immediately visible when deploying or starting the server locally. Previously, missing or invalid configuration would not be detected until the first request hit a handler that called `getAppConfig()`, resulting in a silent failure or a confusing runtime error rather than a clear startup message.

[Jira PP-4124]

## How Has This Been Tested?

- Manual testing in local dev environment, starting the server in the following configurations with valid, invalid, and missing file/unreachable URL:
  - without config (`CONFIG_FILE= npm run start`)
  - with local config file (`CONFIG_FILE=<filename> ...`)
  - with remote config file (`CONFIG_FILE=<url> ...`)
- All checks pass locally.
- [CI checks](https://github.com/ThePalaceProject/web-patron/actions/runs/24896724703) pass.

## Checklist:

### DIFF
--- a/src/instrumentation.ts
+++ b/src/instrumentation.ts
@@ -1,0 +1,12 @@
+export async function register() {
+  if (process.env.NEXT_RUNTIME === "nodejs") {
+    const { getAppConfig } = await import("server/appConfig");
+    try {
+      await getAppConfig();
+    } catch (e) {
+      // Next.js does not exit on register() errors, so we must do it ourselves.
+      console.error(e instanceof Error ? e.message : String(e));
+      process.exit(1);
+    }
+  }
+}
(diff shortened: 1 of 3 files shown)


jonathangreen (approved):
Looks good! Addresses some of the concerns I had in https://github.com/ThePalaceProject/web-patron/pull/125

PR #129, opened 2026-04-28
Update metadata display on book details screen (PP-4048)

PoorBillyPilgrim:
## Description

Add Audience, Language, Narrators, and Duration to the metadata fields displayed on the book details screen, remove audience values from Categories, and reorder all metadata fields to better support patron decision-making.

## Motivation and Context

Fields are ordered by patron decision-making priority. Format, Audience, Category, and Language help patrons quickly determine if a title is relevant and age-appropriate. Narrators and Duration are audiobook-specific fields that influence the commitment decision. Published, Publisher, and Distributor are administrative details least likely to influence a borrow decision. Audience is placed near the top because it's the quickest disqualifier, particularly for parents browsing on behalf of children.

Fields are also now consistent with mobile platforms.

[Jira PP-4048](https://ebce-lyrasis.atlassian.net/browse/PP-4048)

## How Has This Been Tested?

- Tests were added that ensure the new metadata fields are 1) properly extracted from OPDS feed and 2) render expectedly on book details pages.
- Manual testing to confirm metadata rendered as expected

## Checklist:


package-lock.json
tdilauro, line 1: Did you intend to update `package-lock.json`? I didn't see any updates on `package.json`.

src/components/bookDetails/__tests__/BookDetails.test.tsx
```diff
@@ -79,6 +79,23 @@ describe("book details page", () => {
...
+    expect(screen.getByText("Audience:")).toBeInTheDocument();
+  });
+
+  test("shows formatted language", () => {
+    mockSwr({ data: fixtures.book });
+    setup(<BookDetails />);
+    const language = fixtures.book.language as string;
+    expect(language).toBe("en");
+    expect(screen.getByText("English")).toBeInTheDocument();
+    expect(screen.getByText("Audience:")).toBeInTheDocument();
```
tdilauro: "Language" here?


```suggestion
    expect(screen.getByText("Language:")).toBeInTheDocument();
```

src/utils/duration.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * Formats a duration given in seconds as a human-readable string.
+ * Returns "${min} minutes" for durations <= 60 minutes,
+ * otherwise "${hrs} hours, ${min} minutes".
+ */
```
tdilauro, line 5: Just a question: Does the logic here match that of the mobile apps? Since users of CPW will in some cases use the mobile apps, as well, it would good to end up with the same value.

src/utils/__tests__/duration.test.ts
```diff
@@ -0,0 +1,30 @@
+import { expect, test } from "@jest/globals";
+import { formatDuration } from "../duration";
+
+test("returns 0 minutes for 0 seconds", () => {
+  expect(formatDuration(0)).toBe("0 minutes");
+});
+
+test("floors sub-minute durations to 0 minutes", () => {
```
tdilauro: The label is misleading here, as it doesn't match the code or the test result:


```suggestion
test("rounds sub-minute durations up to 1 minute", () => {
```

src/components/bookDetails/__tests__/BookDetails.test.tsx
tdilauro, line 1: Since part of the work here is about the ordering, it would be helpful to add a test that verifies the order of the fields displayed.

tdilauro (approved):
This is a nice improvement! 👏🏽 

A few comments and suggestions below, but I have already approved it.

src/utils/duration.ts
PoorBillyPilgrim: @tdilauro `"${hrs} hours, ${min} minutes"` matches the mobile apps. I'm not sure if `"${min} minutes"` matches. I could check the mobile code bases to be sure.

src/utils/__tests__/duration.test.ts
PoorBillyPilgrim: @tdilauro good catch. I was swapping stuff around and forgot to change.

package-lock.json
PoorBillyPilgrim: @tdilauro jesus thank you. this was not intentional. I'll restore to the previous commit.

src/utils/duration.ts
PoorBillyPilgrim: @tdilauro I don't know swift at all, but it seems that the iOS code shows hour even if its 0: 
```
    private func formatDuration(_ durationInSeconds: String) -> String {
        guard let totalSeconds = Double(durationInSeconds) else {
            return "Invalid input"
        }

        let hours = Int(totalSeconds / 3600)
        let minutes = Int((totalSeconds - Double(hours * 3600)) / 60)

        return String(format: "%d hours, %d minutes", hours, minutes)
    }
   ```

0 hours is likely an edge case as most audiobooks will surely be over an hour long. I could update CPW to just show 0 hours, but I prefer the way I've written it.

tdilauro: Why don't you go ahead with what you've got for now, so that we don't hold up the PR. I'd like to be consistent between clients, but we can check in with the team to make a decision about how this should look.

PR #130, opened 2026-04-28
Unable to navigate to previous page in browser history from login page (PP-4041)

PoorBillyPilgrim:
## Description

Updates how the Next.js Router updates the browser's history stack when redirecting to the login screen based on the user's supported authentication method.

## Motivation and Context

When the `initLogin` function is called within an auth component designed to redirect  to the login page(`Login`, `Catch401`, or `AuthProtectedRoute`), the router previously pushed the derived login URL onto the browser's history stack. This would result in two separate entries in the history stack for the login screen. The login hooks update the URL to accommodate the supported auth type, but this does not visually change what's rendered for the user. So when a user goes backward in the history stack, they remain on the login page. 

Previous behavior
Step | Stack | Behavior
-- | -- | --
User clicks "Sign In" | [home, /login] | /login mounts
Login.tsx effect fires | [home, /login, /login/[methodId] | initLogin(methodId) pushes /login/[methodId]
User hits back | [home, /login] | /login mounts again
Login.tsx effect re-fires | [home, /login, /login/[methodId] | Pushes again, keeps user on login screen

New behavior
Step | Stack | Behavior
-- | -- | --
User clicks "Sign In" | [home, /login] | /login mounts
Login.tsx effect fires | [home, /login/basic] | replace overwrites /login
User hits back | [home] | User goes home

However, views that are reachable without authentication don't need to have their locations replaced within the browser history. `BookDetails` renders a "Borrow" button that redirects to the login screen. 

Previous behavior
Step | Stack with replace | Behavior
-- | -- | --
User on book detail | [home, /book/123] |  
Clicks Borrow without token → initLogin() | [home, /login/basic] | /book/123 is gone
Hits back | [home] | User is incorrectly sent home

To prevent this behavior, non-authenticated views should `push` instead of `replace` because their previous locations are not only meant for redirection. 

Step | Stack with push | Behavior
-- | -- | --
User on book detail | [home, /book/123] |  
Clicks Borrow without token → initLogin() | [home, /book/123, /login/basic] | book details remains on stack
Hits back | [home, /book/123] | User returns to book details

[Jira PP-4041](https://ebce-lyrasis.atlassian.net/browse/PP-4041)

## How Has This Been Tested?

- Updated tests that mock the replace call on the Router
- Manually tested navigating from book details to authenticated routes that utilize `initLogin`
1. `AuthProtectedRoute` -> `/[library]/loans`
2. `Login` -> `/[library]/login`
3. `useCredentials` -> `/[library]/book/[bookUrl]?error={ title, detail}`
4. `Catch401` -> manually thrown `ServerError`

## Checklist:

### DIFF
--- a/src/auth/useLogin.ts
+++ b/src/auth/useLogin.ts
@@ -1,10 +1,13 @@
 import * as React from "react";
-import { useRouter } from "next/router";
+import { useRouter, type NextRouter } from "next/router";
 import { LOGIN_REDIRECT_QUERY_PARAM } from "utils/constants";
 import { UrlObject } from "url";
 import { IS_SERVER } from "utils/env";
 import useLibraryContext from "components/context/LibraryContext";
 import * as _ from "lodash";
+
+/** A navigate function — `router.push` or `router.replace` from `next/router`. */
+type Navigate = NextRouter["push"] | NextRouter["replace"];
 
 export default function useLogin() {
   const { query, replace, asPath, isReady } = useRouter();
@@ -53,11 +56,16 @@ export default function useLogin() {
   // an error can be passed in from a previous login attempt to show on
   // the login screen. It will be passed to the page as a url query param.
   const initLogin = React.useCallback(
-    (methodId?: string, error?: string, useCurrentPageAsRedirect = true) => {
+    (
+      methodId?: string,
+      error?: string,
+      useCurrentPageAsRedirect = true,
+      navigate: Navigate = replace
+    ) => {
       const urlObject = getLoginUrl(methodId, error, useCurrentPageAsRedirect);
       if (!IS_SERVER) {
         // redirect to the login page
-        replace(urlObject, undefined, { shallow: true });
+        navigate(urlObject, undefined, { shallow: true });
       }
     },
     [replace, getLoginUrl]
--- a/src/auth/useLogin.ts
+++ b/src/auth/useLogin.ts
@@ -7,7 +7,7 @@ import useLibraryContext from "components/context/LibraryContext";
 import * as _ from "lodash";
 
 export default function useLogin() {
-  const { query, push, asPath, isReady } = useRouter();
+  const { query, replace, asPath, isReady } = useRouter();
   const { slug } = useLibraryContext();
 
   const getLoginUrl = React.useCallback(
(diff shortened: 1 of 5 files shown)


PoorBillyPilgrim:
@tdilauro Whenever you have a moment, would you be able to test this with the newly added auth methods (SAML, OIDC)? I don't know if the way those methods rely on redirection rely on the browser history acting as it previously did.

src/auth/useLogin.ts
```diff
@@ -57,10 +57,10 @@ export default function useLogin() {
       const urlObject = getLoginUrl(methodId, error, useCurrentPageAsRedirect);
       if (!IS_SERVER) {
         // redirect to the login page
-        push(urlObject, undefined, { shallow: true });
```
tdilauro, line 60: This could be problematic because `initLogin` is used in multiple places where login might be needed, most notably during borrow.

tdilauro:
> @tdilauro Whenever you have a moment, would you be able to test this with the newly added auth methods (SAML, OIDC)? I don't know if the way those methods rely on redirection rely on the browser history acting as it previously did.

This change doesn't fully address issue for all browsers when using the redirect authentication mechanisms, but it materially improves the situation. 🎉

src/auth/useLogin.ts
tdilauro: I just confirmed that there is indeed an issue with borrow. 

Setup: I'm NOT logged in. I pick a book from the catalog and click "Borrow" from the book detail page.

The login dialog is displayed.

If I hit the back button, I would expect to be taken back to the book detail page, but I am taken back to the catalog page.

src/hooks/useBorrow.ts
```diff
@@ -19,7 +21,9 @@ export default function useBorrow(isBorrow: boolean) {
   const borrowOrReserve = async (url: string) => {
     clearError();
     if (!token) {
-      initLogin();
+      // Use push (not initLogin's replace) so the book page is preserved in
+      // history and the back button returns here after sign-in or cancel.
+      router.push(getLoginUrl(), undefined, { shallow: true });
```
tdilauro: Minor - This duplicates some functionality in (and decouples from) `initLogin`, so future us might miss a needed change here if it needs updating there. At minimum, we should add a comment about this on `initLogin`.

tdilauro: Minor - Looks like we could use a test for this behavior, too.

tdilauro (approved):
Looks like this will do the trick. 🚀 

Approved, but a couple suggestions below.

src/hooks/useBorrow.ts
PoorBillyPilgrim: Sorry @tdilauro, this wasn't ready for a review. I just pushed up my intended changes.

tdilauro: Oh, my bad. It already looked like it would get the job done. But what's up there now is better! 🎉

PR #132, opened 2026-05-05
Move Bugsnag / GTM configuration to environment variables (PP-4266)

tdilauro:
## Description

Moves `bugsnag_api_key` and `gtm_id` out of the YAML config file and into environment variables (`BUGSNAG_API_KEY` and `GTM_ID`). The YAML keys are still accepted but emit a deprecation warning and are ignored; their values are never used.

As part of this change, Google Tag Manager (GTM) injection moves from `_app.tsx` (where it previousely read `gtmId` from the app config object passed via `pageProps`) to `_document.tsx` (where `process.env.GTM_ID` is read directly, server-side). The new setup uses the correct placements per GTM's own recommendations — the script tag belongs in `<head>` and the `<noscript>` fallback belongs as the first element in `<body>`.

Additional hardening:
- `GTMScript` and `GTMNoscript` now validate the GTM container ID format (`GTM-[A-Z0-9]+`). A truthy but malformed value causes `GTMScript` to log a `console.error` and render nothing, rather than silently injecting a broken or potentially unsafe script.
- The Bugsnag deprecation warning includes a note when `BUGSNAG_API_KEY` is also absent, so operators know Bugsnag will not be configured (not just that the YAML value is ignored).
- The `GTMScript` component no longer wraps its output in `next/head`, which is unsupported inside `_document.tsx`.

## Motivation and Context

- Keeping the Bugsnag API key and GTM ID in the YAML config file meant they could not be delivered before the first request. For the Bugsnag API key, this meant that reporting could not be available for functions that occurred before the first request. 
- For GTM config, delayed availability meant that we could not locate the GTM-related elements in the recommended locations in the DOM.

[Jira PP-4266]

## How Has This Been Tested?

- Manual testing in my development environment.
- New and updated unit tests cover all changed behavior.
- All checks pass locally.
- CI checks pass.

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,12 +76,17 @@ The app is configured with a YAML file. Point the app at it by setting the `CONF
 | `instance_name` | string | `"Patron Web Catalog"` | Name used in error tracking and debug output. Not patron-facing. |
 | `companion_app` | `"simplye"` \| `"openebooks"` | `"simplye"` | Selects which companion mobile app to reference in redirect prompts. |
 | `show_medium` | boolean | `true` | Whether to display the medium (e-book, audiobook, etc.) label on book cards. |
-| `bugsnag_api_key` | string | — | Bugsnag project API key. Omit to disable error tracking. |
-| `gtm_id` | string | — | Google Tag Manager container ID (e.g. `GTM-XXXX`). Omit to disable analytics. |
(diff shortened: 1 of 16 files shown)


PoorBillyPilgrim (approved):
LGTM 🚀

PR #133, opened 2026-05-07
Prefer heuristic library short name for slug (PP-4263)

tdilauro:
## Description

Adds heuristic derivation of each library's slug from its registry catalog entrie. Extracts the first path component of the catalog root link (e.g. `CA0156` from `https://ca.thepalaceproject.org/CA0156/`), falling back to the same part of the authentication document link, and returning `null` if neither yields a value. `computeSlug` calls it and falls back to the registry identifier as before.

## Motivation and Context

We want to use the library short name as the slug, but we don't maintain that value in the registry (let alone include it in the feed). So, for the time being, we heuristically pull it from the first path component of the library catalog or authentication document URL, since that is a safe place to find it for the Palace Manager library catalog servers in The Palace Project hosted environment. 

[Jira PP-4263]

## How Has This Been Tested?

- New/updated tests to cover the new and changed functionality.
- All checks pass locally.
- [CI checks](https://github.com/ThePalaceProject/web-patron/actions/runs/25460022748) pass.

## Checklist:

### DIFF
--- a/src/server/__tests__/libraryRegistry.test.ts
+++ b/src/server/__tests__/libraryRegistry.test.ts
@@ -97,7 +97,6 @@ function makePagedFeed(
       links: authDocUrl
         ? [
             {
-              rel: "http://opds-spec.org/auth/document",
               type: "application/vnd.opds.authentication.v1.0+json",
               href: authDocUrl
             }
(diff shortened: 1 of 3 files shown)


src/server/__tests__/libraryRegistry.test.ts
```diff
@@ -250,12 +249,12 @@ describe("fetchRegistryLibraries", () => {
       {
         id: "urn:uuid:abc",
         title: "Library A",
-        authDocUrl: "https://a.example.com/auth"
```
tdilauro, line 253: Changed a bunch of these so that the URL doesn't have a path component, since that would result in the first path component becoming the slug.

jonathangreen (approved):
LGTM!

PR #137, opened 2026-05-14
Better handling of cancelled/failed redirect logins (PP-3961)

tdilauro:
## Description

Adds cancel-detection and error-recovery to the SAML, OIDC, and Clever auth redirect handlers, preventing infinite redirect loops when a login attempt is cancelled or fails.

When a patron initiates an external auth login, the handler redirects to an identity provider via a hard browser navigation. Previously, returning to the app without completing auth — via browser back, closing a popup, or an IdP-side failure — caused the handler to immediately redirect again, trapping the patron in a loop.

Key changes:
- **Auth-failure loop fix**: When an IdP redirects back to a protected page with an error parameter, that error is now passed through to the login handler, which displays it as an error message rather than triggering another redirect.
- **Browser-back cancel detection**: sessionStorage flags track the redirect lifecycle. When a user returns to the login page without completing auth, the cancelled message ("Login was cancelled.") is shown with a "Try Again" option rather than immediately re-redirecting.
- **Intentional re-navigation**: When a user on the cancel screen navigates back to the sign-in page (e.g. via the header or the "Try Again" option), the flags are cleared and a fresh redirect proceeds normally.

Special cases:
- **Safari `bfcache` support**: Safari restores pages from its back-forward cache without re-running React effects. A special event listener detects this case so that the cancelled message can be displayed correctly.
- **Cancel button**: A cancel button has been added to allow the user to manually abort and stop the spinner. This is kind of a catch all, but I went down that path because of an issue I ran into with my daily driver ARC Browser:
  - **Popup overlay support (ARC browser)**: ARC's "Little Arc" feature opens external URLs as an in-browser overlay, leaving the main page mounted in spinner state with no standard DOM event to detect closure. An explicit "Cancel" button on the spinner lets the patron manually abort.

## Motivation and Context

Users attempting to log in via redirect-based authentication -- SAML, OIDC, or Clever -- who cancelled mid-flow or failed to authenticate could become trapped in an infinite redirect loop without an obvious way out. 

[Jira PP-3961]

## How Has This Been Tested?

- New unit tests for the improved functionality.
- Manual testing in Firefox, Helium, Safari, and ARC.
- All checks pass locally.
- [CI checks](https://github.com/ThePalaceProject/web-patron/actions/runs/25837388753) pass.

## Checklist:

### DIFF
--- a/src/auth/CleverAuthHandler.tsx
+++ b/src/auth/CleverAuthHandler.tsx
@@ -1,5 +1,6 @@
 import * as React from "react";
-import { OPDS1 } from "interfaces";
+import { useRouter } from "next/router";
+import { ClientCleverMethod } from "interfaces";
 import LoadingIndicator from "components/LoadingIndicator";
 import Stack from "components/Stack";
 import useUser from "components/context/UserContext";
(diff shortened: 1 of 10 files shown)


src/auth/CleverAuthHandler.tsx
PoorBillyPilgrim, line 1: `CleverAuthHandler` and `OidcAuthHandler` are mostly identical save for the flags passed to the `useRedirectCancelDetection` hook. You might consider creating an `ExternalAuthHandler` (or some component to that effect) that received the flags as props. But I can also see the value in keeping these auth handlers totally decoupled in case any changes need to be made to either in the future.

src/auth/useRedirectCancelDetection.ts
PoorBillyPilgrim, line 1: At the moment, the only components that use the `useRedirectCancelDetection` hook are the 3 auth handlers that are wrapped with a `clientOnly` check ensuring this hook is only ever called in the browser. However, if in the future this hook is used in a client component rendered on the server, the `sessionStorage` calls will all throw errors. We might consider wrapping `sessionStorage` with `typeof window !== 'undefined'` to ensure those errors are never thrown.

PoorBillyPilgrim:
Pretty clever solution @tdilauro! I added some comments for review.

src/auth/CleverAuthHandler.tsx
tdilauro: Yeah, I'd like to keep these separate for now for exactly the reason you suggest. I've been trying to have the Clever changes follow the SAML and OIDC ones, but I am not confident at all about them. I just don't want it to be left in a world of its own. And there is still a chance that we will need to make some changes to SAML or OIDC that might cause them to diverge.

src/auth/useRedirectCancelDetection.ts
tdilauro: That's a good point. I will make this change.

tdilauro: I only found one vulnerable spot and added the suggested guard. Thanks!

PoorBillyPilgrim (approved):
👍

PR #138, opened 2026-05-18
Library home button should navigate to catalog home not library website (PP-4354)

PoorBillyPilgrim:
## Description

The "[Library] Home" button now directs the user to the web catalog's home page and the label now reads as "Catalog". Previously, the button would direct users externally to the library home page for that instance of the web catalog.

image

## Motivation and Context

The ambiguous naming of the button label confused users as to the destination of the link. Routing to the catalog home page reinforces the intended destination of the link. This change also aligns with the convention that all the nav links in the navbar link within the application.

RAILs testers using RAILs web catalog instances found that following the link to the RAILs organizational website provided an unnecessarily difficult experience for navigating back to the web catalog.

[Jira PP-4354](https://ebce-lyrasis.atlassian.net/browse/PP-4354)

## How Has This Been Tested?

- Manually tested in Chrome, Firefox, and Safari
- Tested link behavior with mocked Next.js router

## Checklist:

### DIFF
--- a/src/components/Header.tsx
+++ b/src/components/Header.tsx
@@ -63,7 +63,7 @@ const HeaderFC: React.FC<{ className?: string }> = ({ className }) => {
 };
 
 const HeaderLinks: React.FC<{ library: LibraryData }> = ({ library }) => {
-  const { helpWebsite, libraryWebsite } = library.libraryLinks;
+  const { helpWebsite } = library.libraryLinks;
   const libraryName = library.catalogName;
   const { isAuthenticated, isLoading } = useUser();
   const { baseLoginUrl } = useLogin();
@@ -87,6 +87,16 @@ const HeaderLinks: React.FC<{ library: LibraryData }> = ({ library }) => {
           {link.title}
         </AnchorButton>
       ))}
+
+      <NavButton
+        variant="ghost"
+        color="ui.black"
+        href="/"
+        sx={{ whiteSpace: "initial" }}
+      >
+        {`${libraryName} Home`}
+      </NavButton>
+
       {helpWebsite && (
         <AnchorButton
           variant="ghost"
@@ -97,17 +107,7 @@ const HeaderLinks: React.FC<{ library: LibraryData }> = ({ library }) => {
           Help
         </AnchorButton>
       )}
-      {libraryWebsite && (
-        <AnchorButton
-          variant="ghost"
-          color="ui.black"
-          href={libraryWebsite.href}
-          title="help"
-          sx={{ whiteSpace: "initial" }}
-        >
-          {libraryWebsite.title ?? `${libraryName} Home`}
-        </AnchorButton>
-      )}
+
       <NavButton
         variant="ghost"
         color="ui.black"
--- a/src/components/Header.tsx
+++ b/src/components/Header.tsx
@@ -27,7 +27,7 @@ const HeaderFC: React.FC<{ className?: string }> = ({ className }) => {
     >
       <Link
         href="/"
-        aria-label="Library catalog, back to homepage"
+        aria-label="Go to catalog home page"
         sx={{
           display: "flex",
           flexDirection: "column",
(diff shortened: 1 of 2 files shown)


src/components/Header.tsx
```diff
@@ -87,6 +87,16 @@ const HeaderLinks: React.FC<{ library: LibraryData }> = ({ library }) => {
           {link.title}
         </AnchorButton>
       ))}
+
+      <NavButton
```
tdilauro: We've changed the link order (now before the help link) here. Was that intentional?

src/components/__tests__/Layout.test.tsx
```diff
@@ -1,22 +1,55 @@
...
-    setup(<Layout>Child</Layout>);
-    const homeButton = screen.getByLabelText(
-      "Library catalog, back to homepage"
-    );
-
-    // the home button should navigate to "/"
-    expect(homeButton.closest("a")).toHaveAttribute("href", "/testlib");
-  });
+  describe("Navbar", () => {
+    test("Library icon button navigates home", async () => {
```
tdilauro: Minor - The word "home" can be a little confusing here for a couple of reasons:
- In the context of this change, since we were confused about homepage of user's library vs. catalog web app's home page.
- In the context of the web app itself: We have the "root", which is the multi-library link page, and the catalog page of the library.

It would be helpful to make some distinctions here.

src/components/__tests__/Layout.test.tsx
```diff
@@ -1,22 +1,55 @@
...
+      const { user } = setup(<Layout>Child</Layout>);
+      const homeButton = screen.getByLabelText(
+        "Library catalog, back to homepage"
+      );
+
+      // the home button should navigate to "/"
```
tdilauro: Minor - Another area of confusion. We say "... navigate to /", but we're really navigating to the catalog view for the current library. The path below is "/testlib", which is not "/". Confusing.

It would be helpful to clarify what's happening.

src/components/Header.tsx
```diff
@@ -87,6 +87,16 @@ const HeaderLinks: React.FC<{ library: LibraryData }> = ({ library }) => {
...
+        variant="ghost"
+        color="ui.black"
+        href="/"
+        sx={{ whiteSpace: "initial" }}
+      >
+        {`${libraryName} Home`}
```
tdilauro: Minor - It might be more clear to label this as catalog, rather than home. Not sure if we've have any design guidance that would prevent this.


```suggestion
        {`${libraryName} Catalog`}
```
or, perhaps even better, just

```suggestion
        Catalog
```

This shouldn't hold us up, though. We can come back and change this in the future.

tdilauro (approved):
Looks good! 🎸🤘🏽

A question and a few minor suggestions, but approving this now.

src/components/Header.tsx
PoorBillyPilgrim: @tdilauro This was intentional but half-baked. My intention was that a "Home" button is typically the left-most/top-most link in a navbar. Perhaps I also move the "Catalog" button to the left of `library?.headerLinks` as well?

tdilauro: I mainly wanted to make sure that the change was intentional. I would be interested to see what the whole nav bar looks like with all elements populated. But I certainly trust your judgement on what looks and works best here. And we can validate it with some UX review from the team and from users.

src/components/__tests__/Layout.test.tsx
```diff
@@ -1,22 +1,56 @@
...
-    expect(homeButton.closest("a")).toHaveAttribute("href", "/testlib");
-  });
+  describe("Navbar", () => {
+    test("Library icon button navigates to web catalog home", async () => {
+      const { user } = setup(<Layout>Child</Layout>);
+      const libraryIconButton = screen.getByLabelText(
+        "Go to catalog home page"
+      );
+
+      // the home button should navigate to the catalog home page ("/testlib") for t
```
tdilauro: Looks like this comment got cut off.

src/components/__tests__/Layout.test.tsx
tdilauro: Looks like this is sorted out now, aside from the truncated comment mentioned further below.

PR #140, opened 2026-05-22
Reduce excessive registry and auth doc fetches (PP-4342)

tdilauro:
## Description

Two related fixes for redundant network fetches during page generation:

- **Registry cache shared across bundle contexts.** Next.js bundles each page and API route independently, so module-level Maps in `libraryRegistry.ts` were not shared — the instrumentation pre-warm populated one bundle's cache while ISR handlers each held their own empty copy. The fix stores registry state on `global.__cpwRegistryState` so all bundle contexts in the same process share a single cache. The instrumentation hook now also calls `getLibraries()` to pre-warm the registry before any requests are served.

- **Authentication document cache added.** Each ISR route rebuild for the same library (index, loans, book detail, etc.) was fetching the library's authentication document independently. A new server-side cache (also on `global`) stores fetched auth documents with default and configurable TTL (like registry). Concurrent requests for the same URL are coalesced into one in-flight fetch; sequential requests are gated by `refreshMinInterval` (minimum gap between attempts) and `refreshMaxInterval` (staleness threshold), mirroring the semantics already used for registry refresh intervals.

A new `authentication_documents` config section exposes both interval settings. README and `community-config.yml` are updated accordingly.

## Motivation and Context

During ISR page generation, multiple routes are rendered in rapid succession. Without shared caching, each route issued independent fetches to both the library registry and the library's auth document endpoint. This produced redundant network traffic and, in the registry case, could cause spurious (I can use big words, too) 404s on the very first page load when a second simultaneous request saw an empty cache before the first request had finished populating it.

[Jira PP-4342]

## How Has This Been Tested?

- Manual testing and log observation in my local dev environment.
- New unit tests cover the auth document cache: cache hits, TTL-based re-fetch, `refreshMinInterval` throttling, retry after a failed fetch, and coalescing of concurrent requests.
- New instrumentation tests verify that `getLibraries` is called with the loaded config and is skipped when `getAppConfig` fails or when running outside the Node.js runtime.
- All existing new and pre-existing checks pass.
- [CI checks](https://github.com/ThePalaceProject/web-patron/actions/runs/26294427297) pass.

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,7 @@ __To have your library added to the demo, register it with NYPL's Library Regist
   - [Environment Variables](#environment-variables)
   - [Manager, Registry, and Application Configurations](#manager--registry--and-application-configurations)
   - [Libraries and Registries Configuration Settings](#libraries-and-registries-configuration-settings)
+  - [Authentication Document Caching](#authentication-document-caching)
 - [Development](#development)
   - [Contributing](#contributing)
   - [Installing Dependencies](#installing-dependencies)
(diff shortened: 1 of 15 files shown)


PoorBillyPilgrim (approved):
LGTM @tdilauro -- no notes!

PR #141, opened 2026-05-27
Allow users to preview books and audiobooks (PP-4357)

PoorBillyPilgrim:
## Description

image

The mobile apps allow users to preview books and audiobooks. These changes introduce the same functionality into the web catalog. If a preview link is present in an OPDS entry, then the web catalog parses that link for a URL and renders a preview button on the book details page. When a user clicks this button, they are moved to a new tab which renders a third-party web reader.

As is the case with the mobile apps, the button is rendered when a book is

- borrowable
- reservable
- on hold

It should not render once a book has been borrowed.

## Motivation and Context

This change ensures parity with the mobile apps and allows users to sample a title before borrowing.

[Jira PP-4357](https://ebce-lyrasis.atlassian.net/browse/PP-4357)

## How Has This Been Tested?

- Manual testing in Firefox, Chrome, Safari, and Edge (MacOS)
  - preview player opens in new tab (mimicking behavior for external web readers)
  - playback behavior for audiobooks
  - keyboard navigation
- Unit/Integration tests

## Checklist:


src/dataflow/opds1/parse.ts
```diff
@@ -328,6 +332,7 @@ export function entryToBook(entry: OPDSEntry, feedUrl: string): AnyBook {
     url: detailUrl,
     relatedUrl: relatedLink?.href ?? null,
     trackOpenBookUrl: trackOpenBookLink?.href ?? null,
+    previewUrl: previewUrl,
```
tdilauro, line 336: Minor - We could use the short form here:


```suggestion
    previewUrl,
```

src/dataflow/opds1/parse.ts
```diff
@@ -73,6 +73,9 @@ function isRelatedLink(link: OPDSLink) {
 function isTrackOpenBookLink(link: OPDSLink) {
   return link.rel === TrackOpenBookRel;
 }
+function isHtmlPreviewLink(link: OPDSLink) {
+  return link.rel === "preview" && link.type === OPDS1.HTMLMediaType;
```
tdilauro: Minor - Could we use a constant for this rel? I'm a little worried about bare strings and the potential for typos. Using the constant can provide some additional context, as well. It probably belongs with other OPDS constants.

src/components/PreviewButton.tsx
```diff
@@ -0,0 +1,33 @@
...
+      const p = newTab.document.createElement("p");
+      p.textContent = "Loading…";
+      p.style.cssText =
+        "font-family:sans-serif;position:absolute;top:50%;left:50%;transform:translate(-50%,-50%)";
+      newTab.document.body.appendChild(p);
+      newTab.location.href = previewUrl;
+    } else {
+      window.location.href = previewUrl;
+    }
+  }
```
tdilauro: I'm a bit worried about this from a security perspective. The way we're opening this tab, our `_blank` target isn't getting the `rel="noopener,noreferrer"` attributes. Without the "noopener" the new tab would be able to manipulate our page.

Also, I think it would be good to verify that the preview URL we get is an `https` URL.

src/components/CancelOrReturn.tsx
tdilauro, line 1: Now that preview is included in here, we should consider renaming the component (or splitting up?).

src/components/CancelOrReturn.tsx
```diff
@@ -34,23 +37,26 @@ const CancelOrReturn: React.FC<{
...
+          color="ui.gray.light"
+          sx={{
+            bg: "ui.gray.light",
+            color: "ui.gray.dark",
+            "&:hover,&:focus": { color: "ui.gray.dark" }
+          }}
+        >
+          {text}
+        </Button>
+        {previewUrl && <PreviewButton previewUrl={previewUrl} />}
```
tdilauro: There's a potential sneaky little bug here. If the revocation `url` is `null`, we will return early up on l.38 without rendering the this stack at all, so the preview button would not be rendered, even if there were a previewUrl.

tdilauro (commented):
This looks to be in pretty good shape.

A couple of security concerns to address and a couple minor issues to review.

src/components/PreviewButton.tsx
PoorBillyPilgrim: See src/utils/window.ts for my approach to making the tab open more secure.

src/components/CancelOrReturn.tsx
PoorBillyPilgrim: I created a new component (src/components/layouts/FulfillmentButtonStack.tsx) for BorrowOrReserve and CancelOrReturn to share with PreviewButton since they were using the same layout. I moved the error handling into a shared error context so that PreviewButton could also share its own errors.

src/components/CancelOrReturn.tsx
PoorBillyPilgrim: With the new CancelOrReturnOrPreview, I moved the null check on url to within CancelOrReturn so a preview button will still render.

src/components/PreviewButton.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+  function open() {
+    setError(null);
+    try {
+      if (!previewUrl) throw Error();
+      const tab = openPendingTab();
+      tab.navigate(previewUrl);
+    } catch {
+      setError("Error: Could not open preview.");
+    }
```
tdilauro, line 20: We should close `tab`, if we got it here, but we'll need to hoist it out of the try to make it available in the catch, so it might get a little messy if we want to both:
- not open a tab when there's no `previewUrl` and
- keep it a `const` (vs., say a `null`/`undefined` `let`, for which we'd need to type it).

Actually, looking below, the whole component returns `null` if there's no `previewUrl`, so we can probably drop the guard here.


```suggestion
    setError(null);
    const tab = openPendingTab();
    try {
      tab.navigate(previewUrl);
    } catch {
      tab.close();
      setError("Error: Could not open preview.");
    }
```

src/components/bookDetails/__tests__/FulfillmentCard.test.tsx
```diff
@@ -327,12 +327,18 @@ describe("FulfillableBook", () => {
 
   function makeMockTab() {
```
tdilauro: This function with the new changes appears to be duplicated across six files now, so would probably be good to extract it out into `/src/test-utils` (I know, why is this even in `/src`, but that's maybe a question for another PR) before they start to diverge.

src/utils/window.ts
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * Opens a blank tab synchronous so the browser treats it as user-initiated, bypassing popup blockers.
+ * Sets opener=null to prevent tabnabbing and injects a no-referrer meta tag.
+ * Falls back to same-tab navigation when popups are blocked.
+ */
+export function openPendingTab(): PendingTab {
+  const newTab = window.open("about:blank", "_blank");
+  if (newTab) {
+    newTab.opener = null; // breaks reference to window from which tab originated
```
tdilauro, line 22: This seems like it should work!

It's a bit more verbose than `window.open("_blank", "noopener,noreferrer")`, but should get you the behavior you were going for, which would be nice.

tdilauro (approved):
These changes look good to me. 🎸🤘🏽

I left a few more suggestions and unresolved a couple of previous comments, but, there are no show stoppers here, so this is approved, in case you want to get it in this week.

src/utils/window.ts
PoorBillyPilgrim: @tdilauro I tried `window.open("about:blank", "_blank", "noopener,noreferrer")` originally, but it immediately severs a connection with the new tab which prevents us from doing anything with that new tab, like creating elements or updating the href property.

As I understand it, what I've written still allows us to establish an initial connection and then add security measures with `newTab.opener = null` and injecting the no-referrer meta tag in the head all before any actual navigation takes place.

PR #145, opened 2026-06-10
Update visual prominence of external reader buttons (PP-4355)

PoorBillyPilgrim:
## Description

Buttons that open external web readers are now considered as primary actions and are also surfaced on the "My Books" page. The new styling and visual prominence matches that of the iOS and Android apps, further bringing the CPW into branded parity. All other buttons and select other surfaces have also been updated to mirror the visual styling of the apps.

Before
image

After
image

## Motivation and Context

Initial testing highlighted a difficult transition from borrowing to reading. Testers noted that the "Read Online" and "Listen Online" buttons had low visual priority on the page and thus were easily missed by users, especially K-12 students.

[Jira PP-4355](https://ebce-lyrasis.atlassian.net/browse/PP-4355)

## How Has This Been Tested?

- Unit and integration tests have been added to test the new verbiage

## Checklist:

### DIFF
--- a/src/components/BookStatus.tsx
+++ b/src/components/BookStatus.tsx
@@ -1,19 +1,11 @@
 import * as React from "react";
 import { MediumIcon } from "components/MediumIndicator";
 import { AnyBook } from "interfaces";
-import { availabilityString, bookIsFulfillable } from "utils/book";
+import { availabilityString } from "utils/book";
 import { ScreenReaderOnly, Text } from "components/Text";
-import { shouldRedirectToCompanionApp } from "utils/fulfill";
-import SvgPhone from "icons/Phone";
 
 const BookStatus: React.FC<{ book: AnyBook }> = ({ book }) => {
   const { status } = book;
-
-  const redirectUser = bookIsFulfillable(book)
-    ? shouldRedirectToCompanionApp(book.fulfillmentLinks)
-    : false;
-
-  const action = book.format === "Audiobook" ? "Listen" : "Read";
 
   const str =
     status === "borrowable"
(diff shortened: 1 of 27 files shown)


src/icons/Book.tsx
```diff
@@ -1,15 +1,28 @@
...
+//     height="1em"
+//     viewBox="0 0 24 24"
+//     fill="currentColor"
+//     {...props}
+//   >
+//     <path fill="none" d="M0 0h24v24H0V0z" />
+//     <path d="M18 2H6c-1.1 0-2 .9-2 2v16c0 1.1.9 2 2 2h12c1.1 0 2-.9 2-2V4c0-1.1-.9-2-2-2zM6 4h5v8l-2.5-1.5L6 12V4z" />
+//   </svg>
+// );
+
```
tdilauro: We should clear these commented lines out.

src/components/__tests__/BookListItem.test.tsx
```diff
@@ -321,6 +321,7 @@ describe("FulfillableBook", () => {
       availability: undefined
     });
     setup(<BookListItem book={withoutAvailability} />);
-    expect(screen.getByText("Ready to Read!")).toBeInTheDocument();
+    // BookStatus no longer displays this status text
```
tdilauro: Minor - This would be handy as a PR comment, but I don't think it makes as much sense in the code long term. It will lose its context as soon as we merge this PR.

src/components/Button/index.tsx
```diff
@@ -166,7 +166,7 @@ export const InputIconButton = ({
 };
 
 const defaultVariant = "filled";
-const defaultColor = "brand.primary";
+const defaultColor = "ui.black";
```
tdilauro, line 169: This has some downstream repercussions. Since `ui.black` is `#000`, `darken` is not going to change anything. So normal, active, and hover are all going to look the same.

src/components/BookStatus.tsx
```diff
@@ -29,23 +16,19 @@ const BookStatus: React.FC<{ book: AnyBook }> = ({ book }) => {
...
           <MediumIcon book={book} sx={{ mr: 1 }} />
-        )}
-        <Text variant="text.body.bold" sx={{ fontWeight: 600 }}>
-          <ScreenReaderOnly>Book Status: </ScreenReaderOnly>
-          {str}
-        </Text>
-      </div>
+          <Text variant="text.body.bold" sx={{ fontWeight: 600 }}>
+            <ScreenReaderOnly>Book Status: </ScreenReaderOnly>
+            {str}
```
tdilauro: Minor - Given the guard here, it would make sense to change the name of `str` to something like `unfulfillableReason`. That would make its scope an intent more clear. Plus, it'll allow us to get rid of `str`, which was never a great variable name. 😅

tdilauro (approved):
This looks good! 🚀 

A few suggestions below, but I'm going to approve this now.

src/components/Button/index.tsx
PoorBillyPilgrim: @tdilauro great point. I'll mess around with some different hover/active styles.

PR #146, opened 2026-06-10
Support alternate OIDC auth flow via "prompt" param (PP-4429)

tdilauro:
## Description

> [!NOTE]
> The feature described here is basically equivalent to a regular sign in attempt in the absence of [a supporting change in the Palace Manager (PR 3452)](https://github.com/ThePalaceProject/circulation/pull/3452).

This adds a "Use a different account" escape hatch to the OIDC error and cancel screens. Clicking it retries authentication with the `prompt=select_account` query parameter added to the authorization request URL, which instructs the OIDC provider to show its account chooser regardless of any existing session. The normal sign-in flow is unaffected.

image

Also fixes the OIDC redirect URL, now correctly constructing it using `URL`/`searchParams` rather than string concatenation.

## Motivation and Context

When a library uses OIDC for sign-in (e.g. Google SSO), the client auto-redirects to the identity provider. If the user is already signed into a non-permitted account, authentication fails and they land on the error screen with only a "Try Again" button — which retries the same redirect and fails again.

[Jira PP-4429]

## How Has This Been Tested?

- New tests cover the "Use a different account" button appearing in both the error and cancel states, and verify that clicking it navigates to the correct URL with `prompt=select_account`.
- Existing OIDC handler tests updated to reflect the corrected URL format (`?` separator instead of `&`).
- Manual local testing to ensure that the `prompt=select_account` query param makes it through to the Palace Manager.
- All new and existing tests passed.
- CI tests and checks pass.

## Checklist:

### DIFF
--- a/src/auth/OidcAuthHandler.tsx
+++ b/src/auth/OidcAuthHandler.tsx
@@ -26,7 +26,12 @@ const OidcAuthHandler: React.FC<{ method: ClientOidcMethod }> = ({
   const router = useRouter();
   const loginError = extractParam(router.query, "loginError");
 
-  const authUrl = `${method.href}&redirect_uri=${encodeURIComponent(fullSuccessUrl)}`;
+  const authUrl = appendSearchParam(
+    method.href,
+    "redirect_uri",
+    fullSuccessUrl
+  );
+  const altAuthUrl = appendSearchParam(authUrl, "prompt", "select_account");
 
   const { cancelDetected, handleTryAgain, handleCancel } =
     useRedirectCancelDetection({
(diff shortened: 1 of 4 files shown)


src/auth/OidcAuthHandler.tsx
```diff
@@ -65,4 +83,10 @@ const OidcAuthHandler: React.FC<{ method: ClientOidcMethod }> = ({
   );
 };
 
+function appendSearchParam(href: string, name: string, value: string): string {
```
PoorBillyPilgrim, line 86: Are we validating `method.href` before it reaches this component? If not, we probably should. And since `appendSearchParam` is more general, it should probably be validating anyway. We can parse the url using `parseUrl` from `src/utils/parse.ts`. 

```suggestion
function appendSearchParam(href: string, name: string, value: string): string | null {
    const url = parseUrl(href);
    if (!url) return null;
    url.searchParams.set(name, value);
    return url.toString();
}
```

If we add validation, we will need to update how `authUrl` and `altAuthUrl` are created. Probably putting these into a `buildAltAuthUrl` function (or something).

PoorBillyPilgrim (commented):
@tdilauro Looks good! 👍 

I just had a question about the `appendSearchParam` function. I will wait to approve until you've had a chance to offer your own feedback.

src/auth/OidcAuthHandler.tsx
tdilauro: I made `appendSearchParam` as a little helper function for this change. It's not exported or used anywhere else. 

I do agree that we should add a change to validate `method.href` at some point, but I don't think this little helper is the place for it. It should probably happen further upstream on another PR. There is a whole class of these validation issues that need to be handled. I started to introduce that kind of validation with the config file and plan to integrate it everywhere we receive responses. TS is giving us static checking, but we need to add validation to cover runtime.

PoorBillyPilgrim: That all sounds reasonable to me. I'll go ahead and approve.

PoorBillyPilgrim (approved):

PR #151, opened 2026-06-25
Schema validation for registry and authentication document responses (PP-4583)

tdilauro:
## Description

Add runtime validation for library registry feed and authentication document responses using ArkType schemas, complementing the existing compile-time TypeScript types.

- New ArkType schemas in `src/validation/` validate auth document and registry feed responses at the point they are fetched, catching malformed upstream data before it enters the application.
- Schema-derived types for leaf shapes (labels, inputs, localized values, color scheme, features, etc.) replace hand-written interfaces, making the ArkType schemas the single source of truth for those types. Compound types that use CPW-specific narrowing (discriminated auth method union, `AuthDocument` itself) remain hand-written.
- Auth document schema aligned with Palace Manager's `PalaceAuthenticationDocument` Pydantic model, including Palace extension fields (e.g., `features`, `public_key`, `service_description`).
- Registry feed schema aligned with the Library Registry's `OPDSCatalog` output, including `modified` alongside `updated`, nullable `adobe_vendor_id`, `numberOfItems` for paginated feeds, `images` on catalog entries, `properties` on links, and `@type` on facet group metadata.

## Motivation and Context

CPW has TypeScript types enforcing structure at compile time, but no runtime validation of responses from the Circulation Manager (auth documents) or Library Registry (OPDS 2 feeds). A malformed upstream response would silently propagate through the app, potentially causing confusing errors far from the source. Runtime validation catches these at the API boundary and produces clear error messages.

ArkType is already used for config file validation, so this extends the same approach to external API responses. This is foundational work toward making ArkType schemas the single source of truth for related types — deriving TypeScript types directly from the schemas so runtime validation and compile-time type checking stay aligned by default, without the need for manual synchronization.

[Jira PP-4583]

## How Has This Been Tested?

- New test suites in `tests/validation/` cover both schemas: valid documents with all field combinations, rejection of missing required fields, rejection of malformed data (wrong types, null, non-objects).
- Existing tests for `fetchAuthDocument` and `crawlRegistryFeed` updated to use valid response shapes and to verify that invalid responses are rejected with appropriate errors.
- All 803 tests pass across both Jest configurations (JSDOM and Node).
- TypeScript strict-mode typecheck passes.

## Checklist:

### DIFF
--- a/tests/validation/registryFeed.test.ts
+++ b/tests/validation/registryFeed.test.ts
@@ -1,3 +1,4 @@
+/* eslint-disable camelcase */
 /**
  * @jest-environment node
  *
@@ -10,14 +11,20 @@ import { RegistryFeedSchema } from "validation/registryFeed";
 
 function minimal() {
   return {
-    metadata: { title: "Test Registry" },
+    metadata: { title: "Test Registry", adobe_vendor_id: null }, // eslint-disable-line camelcase -- from external API
     links: []
   };
 }
 
 function catalogEntry(id: string, title: string) {
   return {
-    metadata: { id, title, updated: new Date().toISOString(), description: "" },
+    metadata: {
+      id,
+      title,
+      updated: new Date().toISOString(),
+      modified: new Date().toISOString(), // from external API
+      description: ""
+    },
     links: [
       {
         href: `https://${id}.example.com/auth`,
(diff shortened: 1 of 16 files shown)


tests/validation/registryFeed.test.ts
```diff
@@ -0,0 +1,234 @@
...
+  it("rejects when a catalog entry is missing links", () => {
+    const result = RegistryFeedSchema({
+      ...minimal(),
+      catalogs: [
+        {
+          metadata: {
+            id: "x",
+            title: "X",
+            updated: "2024-01-01T00:00:00Z"
+          }
```
PoorBillyPilgrim: Minor suggestion: `metadata` should include `modified` per the `CatalogEntryMetadataSchema` type
```suggestion
            updated: "2024-01-01T00:00:00Z",
            modified: "2024-01-01T00:00:00Z"
          }
```

PoorBillyPilgrim (approved):
Looks great, @tdilauro! I only had a small suggestion. But I consider it non-blocking, so I'm approving the PR.

tests/validation/registryFeed.test.ts
tdilauro: Applied. Thanks!

PR #153, opened 2026-06-30
Multi-library home page live library filtering (PP-4693)

tdilauro:
## Description

- Adds a live-filter input to the multi-library home page. As the patron types, the library is narrowed and sorted by relevance tier and library name using fuzzy character matching with a 200 ms debounce. Matched characters are bolded in each result. When no libraries match the query, a "No libraries match." message replaces the list. 
- For accessibility:
  - A visually-hidden ARIA live region announces the result count (and "best matches first") to screen readers after each debounce cycle.
  - Also fixes a heading hierarchy issue: the "Choose a library:" subheading was `<h3>` and is now `<h2>`.

## Motivation and Context

On deployments with many libraries, the home page presents a long, undifferentiated list that patrons must scroll through manually. Filtering makes it faster and easier to find a specific library by shortening the list.

[Jira PP-4693]

## How Has This Been Tested?

- Manually verified in browser in local dev environment: progressive narrowing, bold highlighting, "No libraries match." message, and native search-input clear button.
- New and updated tests to cover the new functionality.
- All tests and checks pass locally.
- [CI checks](https://github.com/ThePalaceProject/web-patron/actions/runs/28474573760) pass.

## Checklist:

### DIFF
--- a/src/components/MultiLibraryHome.tsx
+++ b/src/components/MultiLibraryHome.tsx
@@ -6,7 +6,11 @@ import useSWR from "swr";
 import { useAppConfig } from "components/context/AppConfigContext";
 import theme from "theme/theme";
 import LibraryHomeLink from "./LibraryHomeLink";
+import TextInput from "components/TextInput";
 import type { ClientLibrary, LibrariesResponse } from "pages/api/libraries";
+
+const FILTER_DEBOUNCE_MS = 200;
+const RESULTS_LIST_ID = "library-filter-results";
 
 async function fetchLibraries(url: string): Promise<LibrariesResponse> {
   const res = await fetch(url);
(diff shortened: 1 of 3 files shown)


src/components/MultiLibraryHome.tsx
```diff
@@ -46,11 +124,45 @@ const MultiLibraryHome: React.FC = () => {
...
+          aria-live="polite"
+          aria-atomic="true"
+          sx={{
+            position: "absolute",
+            width: "1px",
+            height: "1px",
+            overflow: "hidden",
+            clip: "rect(0,0,0,0)",
+            whiteSpace: "nowrap"
+          }}
```
PoorBillyPilgrim, line 210:
```suggestion
          sx={{ variant: "accessibility.visuallyHidden" }}
```
A variant exists in `variants.ts` that covers hiding text. We can reuse it here.

src/components/MultiLibraryHome.tsx
```diff
@@ -6,21 +6,86 @@ import useSWR from "swr";
...
+  let qi = 0;
+  for (let ti = 0; ti < t.length && qi < q.length; ti++) {
+    if (t[ti] === q[qi]) {
+      indices.push(ti);
+      qi++;
+    }
+  }
+  return qi === q.length ? indices : null;
+}
+
```
PoorBillyPilgrim, line 40: An example from Claude on ranking results for search relevancy
```suggestion
type MatchTier = 0 | 1 | 2;
interface LibraryMatch {
  tier: MatchTier;
  indices: number[];
}

/**
 * Classifies how `query` matches `target` and returns the indices to highlight.
 * Prefers contiguous substring matches (word-start > mid-word) over scattered
 * subsequence matches, so results can be ranked by match quality. Returns null
 * when there is no match. An empty query matches everything.
 */
function matchLibrary(query: string, target: string): LibraryMatch | null {
  if (!query) return { tier: 0, indices: [] };
  const q = query.toLowerCase();
  const t = target.toLowerCase();

  const idx = t.indexOf(q);
  if (idx !== -1) {
    const atWordBoundary = idx === 0 || !/[a-z0-9]/.test(t[idx - 1]);
    const indices = Array.from({ length: q.length }, (_, k) => idx + k);
    return { tier: atWordBoundary ? 0 : 1, indices };
  }

  const fuzzy = fuzzyMatchIndices(q, t);
  return fuzzy ? { tier: 2, indices: fuzzy } : null;
}
```

src/components/MultiLibraryHome.tsx
```diff
@@ -35,6 +100,19 @@ const MultiLibraryHome: React.FC = () => {
 
   if (sorted.length === 0) return null;
 
+  const filteredWithIndices = sorted.flatMap(lib => {
+    const displayText = lib.title || lib.slug;
+    const matchIndices = fuzzyMatchIndices(filterQuery, displayText);
+    return matchIndices !== null ? [{ lib, displayText, matchIndices }] : [];
+  });
+
```
PoorBillyPilgrim: This incorporates how to sort the returned matches by relevancy
```suggestion
  const filteredWithIndices = sorted
    .flatMap(lib => {
      const displayText = lib.title || lib.slug;
      const match = matchLibrary(filterQuery, displayText);
      return match
        ? [{ lib, displayText, matchIndices: match.indices, tier: match.tier }]
        : [];
    })
    // Stable sort: alphabetical order (from `sorted`) is preserved within a tier.
    .sort((a, b) => a.tier - b.tier);

```

PoorBillyPilgrim:
@tdilauro This is a great addition. The highlighting is really nice and looks good. I do have a suggestion regarding the relevancy of the search results. For example, the first few results for the search term "illinois" returns results that are not as relevant as those further down the list:
image
I think more relevant search results provides a better UI/UX:
image

I had Claude spin up a quick working example. I've included those snippets as suggestions.

PoorBillyPilgrim (approved):
@tdilauro I've added some comments. I will go ahead and approve because they're non-blocking.

tdilauro:
> @tdilauro I've added some comments. I will go ahead and approve because they're non-blocking.

Thanks for the review, @PoorBillyPilgrim. These are great comments. I had originally stuck to something simpler with the filter to maintain the order, but you make a convincing argument. I'll have a look at that.

PR #155, opened 2026-07-01
Clean up test output for expected errors/warnings (PP-4714)

tdilauro:
## Description

- Adds a shared `expectAndSuppressConsole` test helper that spies on `console.warn`/`console.error`, suppresses messages matching expected patterns, and passes everything else through — so unexpected output (e.g., mismatched error messages or test code errors) remains visible.
- Refactors inline per-test spy setup/teardown into describe-level `beforeEach`/`afterEach` blocks across several test files.
- Patterns at each call site include the full expected message content (including error object strings), serving as inline documentation of what output is expected.

## Motivation and Context

Tests exercising error-handling paths were emitting expected `console.warn`/`console.error` output with multi-line stack traces on every run, making it harder to spot real failures. Blanket suppression was not acceptable because some console output (like React `act()` warnings) reflects genuine test quality issues.

[Jira PP-4714]

## How Has This Been Tested?

Changes are limited to test infrastructure — no production code was modified. The existing test suite validates that affected code paths still produce expected output (spy-call assertions remain where tests verify the warning was actually produced).

## Checklist:

### DIFF
--- a/src/auth/__tests__/BasicTokenAuthHandler.test.tsx
+++ b/src/auth/__tests__/BasicTokenAuthHandler.test.tsx
@@ -14,16 +14,24 @@ import mockAuthenticated, {
   tokenCreds1
 } from "test-utils/mockAuthState";
 import { generateCredentials } from "utils/auth";
+import { expectAndSuppressConsole } from "test-utils/suppressConsole";
 
 beforeEach(() => {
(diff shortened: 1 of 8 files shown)


PoorBillyPilgrim (approved):
This is a great idea! LGTM 🚀

PR #156, opened 2026-07-01
Fix test quality issues (PP-4714)

tdilauro:
## Description

- Replaces `fireEvent.click` with `await user.click` in three tests where the clipboard operation is asynchronous.
- Wraps `jest.advanceTimersByTime` in `await act(...)` in two fake-timer tests.

## Motivation and Context

Several `AccountMenu` tests were producing React `act()` warnings due to async state updates not being properly wrapped. These are real test quality issues, not noise, indicating that component updates were happening after the test framework had relinquished control.
- `fireEvent.click`: the resulting state update was occurring outside of `act`.
-  Advancing timers fires internal state updates need to be inside `act`.

[Jira PP-4714]

## How Has This Been Tested?

- The fix is in the test file itself. 
- All tests pass locally with no `act()` warnings.
- [CI tests](https://github.com/ThePalaceProject/web-patron/actions/runs/28490464588/job/84445827480) pass with no `act()` warnings. 

## Checklist:

### DIFF
--- a/src/components/__tests__/AccountMenu.test.tsx
+++ b/src/components/__tests__/AccountMenu.test.tsx
@@ -1,5 +1,5 @@
 import * as React from "react";
-import { render, waitFor, fireEvent } from "../../test-utils";
+import { render, waitFor, fireEvent, act } from "../../test-utils";
 import { AccountMenu } from "../AccountMenu";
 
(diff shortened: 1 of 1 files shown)


PoorBillyPilgrim (approved):
No notes!

PR #157, opened 2026-07-05
Code review feedback from #153 (PP-4693)

tdilauro:
## Description

Replaces an inline visually-hidden CSS styles block in `MultiLibraryHome` with a shared theme variant (`accessibility.visuallyHidden`).

## Motivation and Context

Mentioned during code review of #153, but I forgot to incorporate it. The inline styles duplicated an existing theme variant rather than reusing it. Eliminates duplication and avoids drift by keeping the styles centralized.

[Jira PP-4693]

## How Has This Been Tested?

No new tests added; this is a style-only refactor with no behavioral change. Existing tests for `MultiLibraryHome` -- and all other tests -- continue to pass.

## Checklist:

### DIFF
--- a/src/components/MultiLibraryHome.tsx
+++ b/src/components/MultiLibraryHome.tsx
@@ -200,14 +200,7 @@ const MultiLibraryHome: React.FC = () => {
           role="status"
           aria-live="polite"
           aria-atomic="true"
-          sx={{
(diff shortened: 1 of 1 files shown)


PoorBillyPilgrim (approved):
🚀

PR #158, opened 2026-07-06
Show library card sign up link when configured for library (PP-4611)

PoorBillyPilgrim:
## Description

image

An online registration link now displays on the sign-in page when offered by a library. When clicked, it opens the user in a new tab at the URL that has been configured for that library in the Palace Manager.

## Motivation and Context

The Palace iOS and Android apps display a "Sign up for a library card" link on a library's sign-in screen whenever a Patron Registration URL has been configured for that library in the Palace Manager (i.e. when a link with `rel="register"` is found within a library's authentication document). This change brings CPW further into parity with the mobile apps.

At present, the Virtual Library Card (VLC) application is the most common -- and currently the only -- registration system behind this URL. The presence of this link should purely be driven by whether a Patron Registration URL is configured for the library, not by the registration system in use. In other words, the authentication method should not determine whether the link renders or not, so the new `SignUpLink` component has been added to the base `LoginWrapper`. 

[Jira PP-4611](https://ebce-lyrasis.atlassian.net/browse/PP-4611)

## How Has This Been Tested?

- Unit tests ensure link only appears when present in a library's authentication document

## Checklist:

### DIFF
--- a/src/auth/LoginWrapper.tsx
+++ b/src/auth/LoginWrapper.tsx
@@ -82,9 +82,7 @@ const LoginWrapper = ({ children }: LoginWrapperProps) => {
           ) : (
             <>
               {children}
-              <div sx={{ textAlign: "center" }}>
-                <SignUpLink />
-              </div>
+              <SignUpLink />
             </>
           )}
         </Stack>
--- a/src/auth/LoginWrapper.tsx
+++ b/src/auth/LoginWrapper.tsx
@@ -8,6 +8,7 @@ import Head from "next/head";
 import BreadcrumbBar from "components/BreadcrumbBar";
 import useLoginRedirectUrl from "auth/useLoginRedirect";
 import { H1, Text } from "components/Text";
+import SignUpLink from "auth/SignUpLink";
 
 /**
  * Redirects on success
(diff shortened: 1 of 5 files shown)


src/auth/LoginWrapper.tsx
```diff
@@ -79,7 +80,12 @@ const LoginWrapper = ({ children }: LoginWrapperProps) => {
               Logging in...
             </Stack>
           ) : (
-            children
+            <>
+              {children}
+              <div sx={{ textAlign: "center" }}>
+                <SignUpLink />
+              </div>
```
tdilauro: Minor - If there is no sign up link, we're going to end up with an empty `div` in here, which will throw off spacing a bit.

We could move the wrapping `div` into `SignUpLink` and only render it after the guard to avoid that.

tdilauro (approved):
Looks good! 🎉 

One small suggestion below.

src/auth/LoginWrapper.tsx
PoorBillyPilgrim: @tdilauro good catch, I'll do that

PR #159, opened 2026-07-07
Display title "Series" metadata with link to series collection search (PP-4649)

PoorBillyPilgrim:
## Description

image

If a title belongs as a part of a series, the name of the series is now listed on the book details page as a link. The link will place users on a collection search page listing all titles within that series.

`opds-feed-parser` expects to find capital-S `schema:Series` which aligns with the legacy NYPL circulation manager. The current Palace Circulation Manager outputs a lowercase-S `schema:series`. The newly added `getSeries` first checks for `schema:series` then `schema:Series` as a fallback.

## Motivation and Context

Listing the series title on the book details page helps users find other titles within that series more easily. The Palace iOS ([PP-4463](https://ebce-lyrasis.atlassian.net/browse/PP-4463)) and Android ([PP-4464](https://ebce-lyrasis.atlassian.net/browse/PP-4464)) apps add a SERIES row to the Book Details Information block, rendering the series name as a link that opens a search for other titles in that series. These changes bring the CPW into further parity with the mobile apps.

[Jira PP-4649](https://ebce-lyrasis.atlassian.net/browse/PP-4649)

## How Has This Been Tested?

- Unit tests added that ensure OPDS feed entries are parsed for series metadata and link surfaces on book details page when present

## Checklist:

[PP-4463]: https://ebce-lyrasis.atlassian.net/browse/PP-4463?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PP-4464]: https://ebce-lyrasis.atlassian.net/browse/PP-4464?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/bookDetails/index.tsx
+++ b/src/components/bookDetails/index.tsx
@@ -23,6 +23,7 @@ import useBreadcrumbContext from "components/context/BreadcrumbContext";
 import { useAppConfig } from "components/context/AppConfigContext";
 import { getAuthors, getLanguageLabel } from "utils/book";
 import Stack from "components/Stack";
+import Link from "components/Link";
 
 export const BookDetails: React.FC = () => {
   const { companionApp, showMedium } = useAppConfig();
@@ -118,6 +119,19 @@ export const BookDetails: React.FC = () => {
               <DetailField heading="Published" details={book.published} />
               <DetailField heading="Publisher" details={book.publisher} />
               <DetailField heading="Distributor" details={book.providerName} />
+              {book.series?.name && book.series?.url && (
+                <DetailField
+                  heading="Series"
+                  details={
+                    <Link
+                      collectionUrl={book.series.url}
+                      sx={{ textDecoration: "underline" }}
+                    >
+                      {book.series.name}
+                    </Link>
+                  }
+                />
+              )}
             </dl>
             <ReportProblem book={book} />
           </div>
--- a/src/components/BookMetaDetail.tsx
+++ b/src/components/BookMetaDetail.tsx
@@ -2,7 +2,7 @@ import * as React from "react";
 
 const DetailField: React.FC<{
   heading: string;
-  details?: string;
+  details?: React.ReactNode;
 }> = ({ heading, details }) =>
   details ? (
     <>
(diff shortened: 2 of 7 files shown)


src/components/bookDetails/index.tsx
```diff
@@ -118,6 +119,19 @@ export const BookDetails: React.FC = () => {
               <DetailField heading="Published" details={book.published} />
               <DetailField heading="Publisher" details={book.publisher} />
               <DetailField heading="Distributor" details={book.providerName} />
+              {book.series?.name && book.series?.url && (
+                <DetailField
+                  heading="Series"
+                  details={
+                    <Link
+                      collectionUrl={book.series.url}
+                      sx={{ textDecoration: "underline" }}
```
tdilauro, line 128: Do we want to completely override `Link`'s base styles here or just override the `text-decoration`?

src/components/bookDetails/index.tsx
```diff
@@ -118,6 +119,19 @@ export const BookDetails: React.FC = () => {
               <DetailField heading="Published" details={book.published} />
               <DetailField heading="Publisher" details={book.publisher} />
               <DetailField heading="Distributor" details={book.providerName} />
+              {book.series?.name && book.series?.url && (
```
tdilauro, line 122: Should we show the series name without the URL if the URL is missing? Seems like it could still provide information to the user, even if they can't link directly to the search. It looks like the parser (`getSeries`) will return a series with an undefined URL when the href is absent or malformed, so it is possible.

tdilauro (approved):
This looks great! 🚀 

A couple of questions below to clarify intentions, but definitely not showstoppers.

src/components/bookDetails/index.tsx
PoorBillyPilgrim: I had the same thought. iOS and android both only display series if name and URL are present, so I was following their lead. This feels like something that can be addressed later.

src/components/bookDetails/index.tsx
PoorBillyPilgrim: This feels more like a one-off than an intentional visual design change for `Link`s. There are also several other `Link` components that have their own styling set on `sx` so this feels like there is a precedent for keeping the base styles of Link lean.

PR #160, opened 2026-07-08
Reduce server-side caching (PP-4753)

tdilauro:
## Description

- Adds a per-request (`getServerSideProps`) variant of the shared page-props wrapper and switches the pages whose dynamic route param is unbounded — book, collection, and the login method page (its param embeds a backend URL or free-form input) — from static generation to server-side rendering, so no cache file is written per unique URL. These responses intentionally keep `Next.js`'s default no-store `Cache-Control` (at least for the time being).
- The remaining statically generated pages now use `fallback: "blocking"` and return `notFound` for unknown library slugs, so junk paths render the custom 404 page without persisting a cache entry on disk.
- The auth document cache keeps repeated renders for the same library from issuing redundant upstream fetches for per-request rendering.

## Motivation and Context

Server-side incremental cache files were using too much storage. Several routes accept arbitrary values in their dynamic params, so any unique URL that was requested (e.g., by crawlers or scanners) persisted a statically generated page to disk, growing the cache without bound.

[Jira PP-4753]

## How Has This Been Tested?

- Manual tested in local dev environment and verified that results for the dynamic routes mentioned above were not cached.
- New and updated tests covering both ISR and SSR wrappers.
- All checks pass locally.
- [CI check](https://github.com/ThePalaceProject/web-patron/actions/runs/28947900932) pass.

## Checklist:

### DIFF
--- a/src/dataflow/__mocks__/withAppProps.ts
+++ b/src/dataflow/__mocks__/withAppProps.ts
@@ -1,8 +1,15 @@
-import { GetStaticProps } from "next";
+import { GetServerSideProps, GetStaticProps } from "next";
 
 export default function withAppProps(pageGetStaticProps?: GetStaticProps) {
   return async (ctx: any) => {
     // Call the inner page getStaticProps directly
     return pageGetStaticProps ? pageGetStaticProps(ctx) : { props: {} };
+  };
+}
+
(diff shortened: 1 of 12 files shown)


jonathangreen (approved):
Looks good!

PR #161, opened 2026-07-09
Parse loans feeds with entries with missing acquisition links (PP-4766)

tdilauro:
## Description

- Classify each feed entry on its own instead of relying on `opds-feed-parser`'s feed-level acquisition-vs-navigation classification: an entry with a complete-entry link is parsed as a book; other entries become navigation links, as before.
- Drop book entries that have no acquisition links at all, matching the current behavior of the iOS and Android mobile apps.

## Motivation and Context

`opds-feed-parser` classifies a feed as an `AcquisitionFeed` only when **every** entry has an acquisition link (aligned with OPDS 1.2  section2); otherwise the whole feed becomes a `NavigationFeed`. A loans feed containing even a single title with no acquisition links (e.g., a book borrowed in another system with a format unsupported in Palace) was therefore classified as a navigation feed and every book in it was turned into a navigation link (currently unhandled). So, the patron's My Books page showed the empty-bookshelf message even though the feed looked valid.

[Jira PP-4766]

## How Has This Been Tested?

- Manual testing in my local dev environment verifying that all holds and loans with acquisition links were displayed on the My Books page of the patron who was experiencing this issue.
- New unit test to verify the new behavior.
- All checks pass locally.
- [CI checks](https://github.com/ThePalaceProject/web-patron/actions/runs/29046167591) pass.

## Checklist:

### DIFF
--- a/src/dataflow/opds1/__tests__/parse.test.ts
+++ b/src/dataflow/opds1/__tests__/parse.test.ts
@@ -847,6 +847,46 @@ test("extracts navigation link info", () => {
   expect(link.url).toBe(`http://test-url.com/${navigationLink.href}`);
 });
 
+/**
+ * opds-feed-parser produces a NavigationFeed whenever any entry lacks
+ * acquisition links; books must still be extracted from it, and the
+ * link-less entries dropped.
+ */
+test("extracts books regardless of feed type", () => {
+  mockConfig();
+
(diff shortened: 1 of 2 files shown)


PoorBillyPilgrim (approved):
LGTM 🚀

PR #163, opened 2026-07-16
Standardize breadcrumbs for search results collection (PP-4675)

PoorBillyPilgrim:
## Description

When computing breadcrumbs, the CPW will now consider whether or not the collection is a search results feed. When that is the case, the breadcrumbs will now only display the name provided for the catalog root and "Search." For example, a search conduced from a "New & Noteworthy" collection lane will have its name dropped:

`All Books / New & Noteworthy / Search` -> `All Books / Search`

## Motivation and Context

When conducting a search from any page, the breadcrumbs previously included the lane from where the search originated. This incorrectly suggested that the search would be scoped to items within that collection. Searches from any page are performed on the entire catalog. Displaying the breadcrumbs as Catalog Root / "Search" provides a more accurate description of the search results to the user.

[Jira PP-4675](https://ebce-lyrasis.atlassian.net/browse/PP-4675)

## How Has This Been Tested?

- Unit tests check for presence of lane when running `computeBreadcrumbs`

## Checklist:

### DIFF
--- a/src/computeBreadcrumbs.tsx
+++ b/src/computeBreadcrumbs.tsx
@@ -14,8 +14,39 @@ const urlComparator = (
   return !!(url1 && url2) && url1 === url2;
 };
 
+/**
+ * A collection is a search results feed when its own url
+ * shares a pathname with its search description url
+ */
+export function isSearchResultsCollection(
+  collection?: CollectionData
+): boolean {
+  if (!collection?.url || !collection.searchDataUrl) return false;
+  // provide dummy base URL in case collection.url
+  // or collection.searchDataUrl are relative links
+  const base = "http://base";
+  try {
+    return (
+      new URL(collection.url, base).pathname ===
+      new URL(collection.searchDataUrl, base).pathname
+    );
+  } catch {
+    return false;
+  }
+}
+
 const computeBreadcrumbs = (collection?: CollectionData): LinkData[] => {
   let links: LinkData[] = [];
+
+  // Currently, search runs across the entire catalog.
+  // So no matter where search originates (home, lane, my books, etc.),
+  // the breadcrumbs should be uniform: Catalog Root / Search
+  if (collection && isSearchResultsCollection(collection)) {
+    return hierarchyComputeBreadcrumbs(
+      { ...collection, parentLink: undefined },
+      urlComparator
+    );
+  }
 
   if (
     collection &&
--- a/src/computeBreadcrumbs.tsx
+++ b/src/computeBreadcrumbs.tsx
@@ -18,18 +18,15 @@ const urlComparator = (
  * A collection is a search results feed when its own url
  * shares a pathname with its search description url
  */
-export function isSearchResultsCollection(
-  collection?: CollectionData
-): boolean {
-  if (!collection?.url || !collection.searchDataUrl) return false;
-  // provide dummy base URL in case collection.url
-  // or collection.searchDataUrl are relative links
-  const base = "http://base";
+function isSearchResultsCollection(collection?: CollectionData): boolean {
+  if (!collection?.url || !collection?.searchDataUrl) return false;
+  // Palace CM currently always provides absolute collection and search URLs
   try {
-    return (
-      new URL(collection.url, base).pathname ===
-      new URL(collection.searchDataUrl, base).pathname
+    const { pathname: collectionPathname } = new URL(collection.url);
+    const { pathname: searchDataUrlPathname } = new URL(
+      collection.searchDataUrl
     );
+    return urlComparator(collectionPathname, searchDataUrlPathname);
   } catch {
     return false;
   }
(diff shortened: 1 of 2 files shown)


src/computeBreadcrumbs.tsx
```diff
@@ -14,9 +14,40 @@ const urlComparator = (
...
+  }
+}
+
 const computeBreadcrumbs = (collection?: CollectionData): LinkData[] => {
   let links: LinkData[] = [];
 
+  // Currently, search runs across the entire catalog.
+  // So no matter where search originates (home, lane, my books, etc.),
+  // the breadcrumbs should be uniform: Catalog Root / Search
+  if (collection && isSearchResultsCollection(collection)) {
```
tdilauro, line 41: Minor - Can probably drop the `collection` guard here, since calling `isSearchResultsCollection` with a falsy `collection` returns false anyway. 


```suggestion
  if (isSearchResultsCollection(collection)) {
```

src/computeBreadcrumbs.tsx
```diff
@@ -14,9 +14,40 @@ const urlComparator = (
   return !!(url1 && url2) && url1 === url2;
 };
 
+/**
+ * A collection is a search results feed when its own url
+ * shares a pathname with its search description url
+ */
+export function isSearchResultsCollection(
```
tdilauro: Do we need this export? I don't see it getting imported anywhere.

src/computeBreadcrumbs.tsx
```diff
@@ -14,9 +14,40 @@ const urlComparator = (
...
+  collection?: CollectionData
+): boolean {
+  if (!collection?.url || !collection.searchDataUrl) return false;
+  // provide dummy base URL in case collection.url
+  // or collection.searchDataUrl are relative links
+  const base = "http://base";
+  try {
+    return (
+      new URL(collection.url, base).pathname ===
+      new URL(collection.searchDataUrl, base).pathname
```
tdilauro: Do we need to normalize the trailing slashes. I don't see any test cases that cover that.

Seems like we could use `urlComparator` (above), since it handles that case.

src/computeBreadcrumbs.tsx
```diff
@@ -14,9 +14,40 @@ const urlComparator = (
...
+    );
```
tdilauro: URL comparison is really tricky. We could really use more test coverage here for (1) trailing slash mismatch cases and (2) absolute, absolute-path, and relative-path URL forms. I'd suggest parameterized tests for these. Otherwise the tests become tedious to write and it becomes non-obvious that the same checks are being run in each case.

tdilauro (commented):
This looks good overall, but I have some concerns, mostly focused on the URL comparison and associated test coverage.

src/computeBreadcrumbs.tsx
PoorBillyPilgrim: @tdilauro This is good feedback, thank you.

I almost reached out to you about how best to identify a search results feed. I was basing the shape of the URLs off of what I could see on feeds from Main Street and A1QA. I didn't love comparing the URLs but it seemed it was the most reliable option.

src/computeBreadcrumbs.tsx
PoorBillyPilgrim: For a similar reason below: I'm not sure about the best way to identify a search results feed. I wasn't sure if the query parameters would match and therefore opted to compare on the pathname only.

src/computeBreadcrumbs.tsx
tdilauro: I think doing URL comparison is a fine approach. It's just tricky, since there are a bunch of cases. That's why I suggested having more tests, since we can define what _should_ happen in those test cases to ensure that we get the behavior we want. I suggested parameterized cases because it removes the disincentive to cover all of the cases that matter.

tdilauro: Also, the best base URL is probably that of the feed or whatever other response from which you's getting the relative URL.

PoorBillyPilgrim: Thanks @tdilauro, this is helpful. Do you have a sense on whether I should just be matching on pathname or the entire URL?

tdilauro: I took a look at the `circulation` code to see.

- Expect absolute URLs (at least for Palace CMs):
  - For a search results feed, collection `url` and `searchDataUrl` should be identical in scheme, host, and path, differing only in the query string ("q", plus pagination). Both are **always absolute** (no relative-path hrefs). This is assuming that we'll only ever be fronting Palace Managers and not other OPDS server implementations, but that seems like a good place to start.

- Normalized path by removing trailing slash:\
  - As I suspected, though, the trailing slash can be an issue. The root search is registered via `library_dir_route("/search", ...)`, which registers both `/search` and `/search/` with `strict_slashes=False`, meaning a PM serves either with a 200 (deliberately w/o redirect for CORS). `url_for` emits `/search/` (with trailing slash) for root, but `/search/<lane-id>` (no trailing slash) for lanes.

Hope that helps.

PoorBillyPilgrim: @tdilauro this is immensely helpful. 🫡

tdilauro (approved):
This looks good! 🎉

PR #164, opened 2026-07-17
Experimental canonical resource URL support POC (PP-3760)

tdilauro:
## Description

- Adds an experimental cross-library item landing page at `/{landing-slug}/{workId}` (default slug: `_item_`), giving each work a canonical URL that is not tied to any one library. The page lets the patron choose a library (with type-ahead filtering), confirms the item is available there, and then opens the item's book page in that library's catalog.
- Deployments with a single library skip the selector and redirect straight to the item's book page.
- The landing slugs are configurable at runtime via a new `item_landing_slugs` config option: multiple item-level resource slugs can be active at once (mainly to support migrations from one to another), an empty list disables the page, entries colliding with reserved Next.js path segments are rejected, and a library whose slug collides with a landing slug is kept but logged as shadowed.
- Adds a `GET /api/catalog-url` endpoint that resolves a library's OPDS catalog root from its authentication document (served from the existing in-process cache).
- Selector resilience: selection is blocked while a lookup is in flight, lookups are bounded by a 15-second timeout and can be canceled with Escape, and all failure states (library unreachable, library unavailable, item not in collection, navigation failure) share one dismissible error panel with "Back to search" and "OK" actions that manage focus.

## Motivation and Context

Proof of concept. OPDS entry URLs are library-specific, so there has been no stable way to link to an item without already knowing the patron's library. A canonical item-level URL can be shared or referenced externally. The landing page bridges it to whichever library the patron uses.

[Jira PP-3760]

## How Has This Been Tested?

- New / updated tests for the new functionality.
- All tests and checks pass locally.
- [CI checks / build](https://github.com/ThePalaceProject/web-patron/actions/runs/29600814227) pass.

## Checklist:

### DIFF
--- a/src/components/WorkLibrarySelector.tsx
+++ b/src/components/WorkLibrarySelector.tsx
@@ -1,0 +1,284 @@
+import * as React from "react";
+import useSWR from "swr";
+import { useRouter } from "next/router";
+import type { ClientLibrary, LibrariesResponse } from "pages/api/libraries";
+import type { CatalogUrlResponse } from "pages/api/catalog-url";
+import LibraryFilterList from "components/LibraryFilterList";
+import { fetchLibraries } from "dataflow/fetchLibraries";
+import { buildWorkUrl } from "utils/workUrl";
+
+/*
+ * Bounds a selection resolve (catalog URL lookup plus availability pre-flight)
+ * so a hung server cannot leave every card blocked. On timeout the in-flight
+ * fetch aborts and the generic connectivity error is shown.
+ */
+export const RESOLVE_TIMEOUT_MS = 15000;
+
+interface WorkLibrarySelectorProps {
+  workId: string;
+}
+
+const WorkLibrarySelector: React.FC<WorkLibrarySelectorProps> = ({
+  workId
+}) => {
+  const { data, error } = useSWR<LibrariesResponse>(
+    "/api/libraries",
+    fetchLibraries
+  );
+  const searchContainerRef = React.useRef<HTMLDivElement>(null);
+  /*
+   * Slug of the card currently resolving a selection, or null. Held here
+   * rather than per card so a resolve in flight blocks selection on every
+   * card; otherwise two cards could race and the later navigation would win.
+   */
+  const [resolvingSlug, setResolvingSlug] = React.useState<string | null>(null);
+
+  function focusSearch() {
+    const input = searchContainerRef.current?.querySelector("input");
+    input?.focus();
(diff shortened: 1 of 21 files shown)


src/components/WorkLibrarySelector.tsx
```diff
@@ -0,0 +1,284 @@
...
+    const input = searchContainerRef.current?.querySelector("input");
+    input?.focus();
+    searchContainerRef.current?.scrollIntoView({
+      behavior: "smooth",
+      block: "nearest"
+    });
+  }
+
+  if (error) return <p>Unable to load the library list.</p>;
+  if (!data?.libraries) return null;
```
PoorBillyPilgrim, line 46: UX question -- should we render "No libraries available" if none are returned instead of rendering nothing? At the moment, if no libraries are returned, the page just renders a "Find a Library" heading.

src/components/WorkLibrarySelector.tsx
```diff
@@ -0,0 +1,284 @@
...
+          return (
+            <LibrarySelectorCard
+              library={library}
+              workId={workId}
+              label={label}
+              highlighted={highlighted}
+              onBackToSearch={focusSearch}
+              resolvingSlug={resolvingSlug}
+              onResolvingChange={setResolvingSlug}
+            />
```
PoorBillyPilgrim, line 76: Minor -- the filter input stays active while a slug is being resolved. If a user updates the search keyword while the resolve is in flight and the search updates before the slug is resolved _and_ the library slug being resolved is removed from the filter list _and_ the resulting library does not have that item, the error card is unmounted.

This is pretty minor and an edge case but something I discovered while manually testing. You could pass `disabled={resolvingSlug !== null` down to the filter input within `LibrarySelectorCard`, but I don't think it's necessary. Up to you.

PoorBillyPilgrim (approved):
This is a cool addition. Worked seamlessly while manually testing. I've added a couple notes, neither are show-stopping by any means.

I'm gonna go ahead and approve.

src/components/WorkLibrarySelector.tsx
tdilauro: Good call! I got that pattern from the multi-library home page, so I'll fix that one, too.

src/components/WorkLibrarySelector.tsx
tdilauro: Unfortunately, `disable`ing could drop keyboard focus in some browsers, so I used `aria-disabled`. And, as you said, it's quite a race to trigger this. I'll leave this one for now.

PR #165, opened 2026-07-18
Support alternate SAML auth flow via the "force_authn=true" param (PP-4792)

tdilauro:
## Description

Adds a "Use a different account" option to the SAML login error and cancelled-login screens, the SAML analogue of the OIDC account switching added in #146. Clicking it sends the patron to the Palace Manager login URL with `force_authn=true`, which the PM translates into `ForceAuthn` on the SAML AuthnRequest so the IdP re-authenticates the patron instead of reusing an existing session.

Along the way:
- The external-redirect login flow (redirect, cancel detection, error/cancel screens, account switching) is now shared between the SAML and OIDC handlers instead of duplicated. Each handler supplies only its storage keys and its provider-specific "switch account" query parameter.
- Auth URLs are now built via a shared helper.
- Fixes a bug in the OIDC flow (from #146) in which switching accounts cleared the redirect flag before leaving the app, so backing out of the provider silently redirected the patron to the provider again instead of showing the cancel screen.

## Motivation and Context

When a SAML login fails (for example, the IdP session belongs to an account with no patron record, or a shared device holds someone else's session), retrying just reuses the same IdP session and fails the same way. Patrons need a way to force the IdP to re-authenticate so they can sign in with a different account. The Palace Manager side of this was added in ThePalaceProject/circulation#3574.

[Jira PP-4792]

## How Has This Been Tested?

- New tests cover the "Use a different account" button in both the error and cancelled states for SAML and OIDC. The redirect URL is parsed and checked by host and query parameter; and the redirect/cancel flags are verified to be in the right state on the way out so backing out of the IdP is detected as a cancel.
- All checks / build pass locally.
- [CI checks / build](https://github.com/ThePalaceProject/web-patron/actions/runs/29627225847) pass.

## Checklist:

### DIFF
--- a/src/auth/OidcAuthHandler.tsx
+++ b/src/auth/OidcAuthHandler.tsx
@@ -1,18 +1,16 @@
 import * as React from "react";
-import { useRouter } from "next/router";
 import { ClientOidcMethod } from "interfaces";
-import LoadingIndicator from "components/LoadingIndicator";
-import Stack from "components/Stack";
-import useUser from "components/context/UserContext";
-import useLoginRedirectUrl from "auth/useLoginRedirect";
 import { clientOnly } from "components/ClientOnly";
-import extractParam from "dataflow/utils";
-import Button from "components/Button";
-import { useRedirectCancelDetection } from "auth/useRedirectCancelDetection";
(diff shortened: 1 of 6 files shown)


PoorBillyPilgrim (approved):
LGTM 👍

PR #166, opened 2026-07-21
OPDS 2 support MVP (PP-2718)

tdilauro:
## Description

Adds OPDS 2 (JSON) support as a format-agnostic alternative to the existing OPDS 1.x (Atom/XML) catalog client, gated behind a feature flag.
- New format-agnostic fetch layer (`dataflow/catalog.ts`) negotiates OPDS 2 via an `Accept` header and dispatches to an OPDS 2 or OPDS 1.x parser based on the response's `Content-Type`. Servers that don't serve OPDS 2 (or ignore the header) fall back to the existing (and unchanged) Atom parser. All call sites that previously imported `fetchCollection`/`fetchBook` from `dataflow/opds1/fetch` now import them from here instead.
- New OPDS 2 catalog parser and lenient runtime validation (`dataflow/opds2/`, `validation/opds2Catalog.ts`) map Palace Manager-produced OPDS 2 JSON into the same internal `Book`/`CollectionData` model the OPDS 1 parser produces, including borrow/hold/loan status, fulfillment and open-access links (with indirection chains, e.g. bearer-token-wrapped or streaming formats), series, and facets/groups/navigation. Schema mismatches are reported as telemetry only and never block rendering.
- Feature flag env var `PALACE_CPW_FEATURE_OPDS2` controls whether OPDS 2 negotiation is attempted; off by default.

**Supporting changes**: 
- OPDS-entry indirection types are normalized between the two spec's differing media-type spellings.
- `getMedium` now recognizes OPDS 2's `@type` metadata.

**Note for review**: 
- `jest.config.js` now sets `restoreMocks: true` so spies auto-restore after each test, fixing a latent footgun that I noticed after forgetting to manually `mockRestore()`. It could leak a mock into later tests. It is unrelated to OPDS 2 but was needed to keep the new OPDS 2 test suites isolated from each other, and it required moving a handful of pre-existing module-scope spies into `beforeEach` in files that I wouldn't have otherwise touched: `
- test-utils/mockToDateString.ts`, 
- `test-utils/index.tsx`, 
- `UserContext.test.tsx`, 
- `ContextProvider.test.tsx`, and
- `BookDetails.test.tsx`.
No test behavior changes here, so it's probably safe to just skim these.

## Motivation and Context

This adds client support so the web patron app can consume OPDS 2 feeds, working toward parity with the mobile apps and what the circulation manager exposes, without disrupting existing OPDS 1.x deployments.

[Jira PP-2718]

## How Has This Been Tested?

- Manual testing in my local development environment.
- New and updated tests for the new functionality.
- Tests and checks pass locally.
- [CI tests / checks](https://github.com/ThePalaceProject/web-patron/actions/runs/29867770774) pass.

## Checklist:

### DIFF
--- a/src/dataflow/catalog.ts
+++ b/src/dataflow/catalog.ts
@@ -1,0 +1,158 @@
+import { OPDSEntry, OPDSFeed } from "opds-feed-parser";
+import ApplicationError, { ServerError } from "errors";
+import { AnyBook, CollectionData, OPDS2 } from "interfaces";
+import fetchWithHeaders from "dataflow/fetch";
+import {
+  fetchBook as fetchOpds1Book,
+  fetchCollection as fetchOpds1Collection,
+  parseOPDSText
+} from "dataflow/opds1/fetch";
+import {
+  entryToBook,
+  feedToCollection as opds1FeedToCollection
+} from "dataflow/opds1/parse";
+import {
+  feedToCollection as opds2FeedToCollection,
+  publicationToBook
+} from "dataflow/opds2/parse";
+import {
+  lenientValidate,
+  Opds2FeedSchema,
+  Opds2PublicationSchema
+} from "validation/opds2Catalog";
+
+/**
+ * Format-agnostic catalog fetching. When OPDS 2 is enabled, requests are
+ * made with an Accept header preferring OPDS 2 (JSON) and the response
+ * Content-Type decides which parser handles the body; servers that do not
+ * serve OPDS 2 (including any that ignore the Accept header) fall back to
+ * the OPDS 1.x (Atom) parser. When disabled, requests are delegated
+ * unchanged to the OPDS 1.x fetchers.
+ */
+
+// Quality-ranked so that a Palace Circulation Manager negotiates OPDS 2
+// while servers without OPDS 2 support still get an Accept value they can
+// satisfy.
+export const OPDS2_FEED_ACCEPT_HEADER = `${OPDS2.BaseDocumentMediaType}, application/atom+xml;q=0.9, */*;q=0.1`;
+export const OPDS2_ENTRY_ACCEPT_HEADER = `${OPDS2.PublicationMediaType}, application/atom+xml;q=0.9, */*;q=0.1`;
+
+let _opds2Enabled = false;
+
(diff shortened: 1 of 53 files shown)


src/dataflow/catalog.ts
```diff
@@ -0,0 +1,166 @@
...
+export function setOpds2Enabled(enabled: boolean): void {
+  _opds2Enabled = enabled;
+}
+
+// Each fetcher matches only the OPDS 2 media types it can handle; anything
+// else falls through to the Atom parser, which fails with a clear error
+// instead of leniently mis-parsing the wrong document kind.
+function isOpds2ContentType(contentType: string, expected: string[]): boolean {
+  return expected.some(type => contentType.includes(type));
+}
```
PoorBillyPilgrim: The function body and its parameter types are too general given its name and its specific use cases. Could we maybe narrow the typing for the `expected` parameter with a union of the newly added`OPDS2.*` media types?

src/dataflow/opds2/parse.ts
```diff
@@ -0,0 +1,647 @@
...
+  publications: Opds2Publication[],
+  feedUrl: string
+): AnyBook[] {
+  return publications
+    .filter(pub => asArray(pub.links).some(isAcquisitionLink))
+    .map(pub => publicationToBook(pub, feedUrl));
+}
+
+function groupToLane(group: Opds2Group, feedUrl: string): LaneData | null {
+  const publications = group.publications ?? [];
```
PoorBillyPilgrim: Should `group.publications` be wrapped with `asArray`? `feed.publications` at line 564 should also be wrapped by`asArray` (at least I think). I just noticed that all the other sibling props on `Opds2FeedSchema` were arrays. Could possibly throw a runtime error if we filter or map on that elsewhere.

PoorBillyPilgrim (approved):
This looks great @tdilauro. Glad we're introducing OPDS 2 support. I've added a few comments but approving.

src/dataflow/opds2/parse.ts
tdilauro: Yes. Good catch! Will fix.

tdilauro: Fixed and added tests to cover.

PR #173, opened 2026-08-03
Only render user account header items in the client

PoorBillyPilgrim:
## Description

Wraps user account header components in `ClientOnly` component, ensuring they are only rendered in the client.

No changes in functionality are included.

## Motivation and Context

image

The hooks that check for user credentials reference local storage. The Web Storage API is not available on the server, so the server always sends the "Sign In" link as an `<a>` element. This causes a hydration error when a user is already "signed in," i.e. have credentials present in local storage, because React expects to find a `<button>` element.

## How Has This Been Tested?

- All existing tests pass

## Checklist:


tdilauro (approved):
LGTM! 🎸🤘🏽

PR #174, opened 2026-08-04
Introduce internationalization (i18n) support and tooling (PP-4708)

PoorBillyPilgrim:
## Description

Introduces internationalization (i18n) support and tooling allowing patrons to navigate the application in English, French, Italian, and Spanish, with English set as the default language.

These changes use Next's built-in locale strategy - [Sub-path Routing](https://nextjs.org/docs/pages/guides/internationalization#sub-path-routing) - to route pages through translations. For example, the French translations for`https://example.com/book` can be found at `https://example.com/fr/book`. 

The application adds dependencies from the `i18next` ecosystem (`i18next`, `next-i18next`, `react-i18next`, `i18next-cli`) and new configuration files (`next-i18next.config.js` and `i18next.config.ts`) to help manage, update, and sync translations. Notably, new scripts have been added that provide status updates on the current state of translations as well tooling to enable quick translation key syncing between the default language ("English") and any other locale files found in `/public/locales/`. Custom linting rules have also been added to `.eslintrc.js` to catch hardcoded strings and prompt devs to add translations when available.

A language switcher has been added to the app's header. Selecting a new language will reload the app with new translations. The language switcher will render the available languages in the current locale's translation (e.g. selecting "French" changes the options to "Anglais", "Français", "Italien", and "Espagnol") along with the ISO 639 language code for each to aid in language selection when landing on a language not belonging to the user.

> __Notes on Future Improvements:__ 
> - This PR only adds support for translations; the translations themselves will be added in subsequent pull requests.
> - Automatic language detection will also be put off onto another pull request. Next.js offers automatic detection from the values found in `Accept-Language` Headers. However, this could be problematic as that might filter out Ebooks and Audiobooks that do not match the locale passed in the Header, based on comments in the code. Other options exist, such as [i18next-browser-languageDetector](https://github.com/i18next/i18next-browser-languageDetector), that will require more hands-on wiring. 
> - The default language is set as English, but this could be easily overridden as an App config, allowing for libraries to determine if they want the default to be Italian, for example.

## Motivation and Context

Brings `web-patron` further into parity with mobile apps. These changes also expand the CPW's target audience to allow for native Spanish, French, and Italian speakers, paving the way for international library use.

These changes also borrow heavily from the i18n implementation undertaken by [`ekirjasto-web-patron`](https://github.com/NatLibFi/ekirjasto-web-patron).

[Jira PP-4708](https://ebce-lyrasis.atlassian.net/browse/PP-4708)

## How Has This Been Tested?

- Manually tested on mobile (iOS)
- New and existing tests pass
- All CI checks pass

## Checklist:


README.md
tdilauro, line 1: There seem to be a bunch of unrelated changes in this file and elsewhere. Looks like it might be overaggressive linting / prettifying.

src/components/Header.tsx
```diff
@@ -122,10 +122,12 @@ const HeaderLinks: React.FC<{ library: LibraryData }> = ({ library }) => {
       ) : isLoading ? (
         <Button loading />
       ) : (
-        <NavButton color="ui.black" href={baseLoginUrl}>
+        <NavButton color="ui.black" href={baseLoginUrl} sx={{ mr: 3 }}>
```
tdilauro: Is this the right place for the margin? The spacing would disappear in the authenticated and loading states above. Could we put a left margin on the `<LanguageSelector />`?

src/components/LanguageSelector.tsx
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    // change the language:
+    // - update the URL and reload the page with the selected language
+    router
+      // try to push the same path to the router with a new locale
+      .push(router.asPath, router.asPath, { locale: language })
+      .catch(error => {
+        // if navigation fails, just console log error
+        console.error("Failed to change language", error);
```
tdilauro: We should probably handle this in our error tracking (`analytics/track`).

tsconfig.json
```diff
@@ -39,6 +39,8 @@
     "**/*.tsx",
     "next.config.js",
     "public/sw.js",
-    "src/icons/_template.js"
+    "src/config/fetch-config.js",
+    "src/icons/_template.js",
+    "next-i18next.config.js"
   ]
-}
+}
```
tdilauro: Missing newline at EOF.

i18next.config.ts
```diff
@@ -0,0 +1,112 @@
...
+
+// define the default namespace (file) for translations as "translations"
+const defaultTranslationNamespace = "translations";
+
+// define the input file patterns to search for translation keys
+// We include all .tsx and .jsx files in the components and pages directories
+const inputFiles = ["src/{components,pages}/**/*.{tsx,jsx}"];
+
+// define the output path to public/locales directory,
+// and organise the translation files by language and namespace,
```
tdilauro: Minor - typo: "organise" -> "organize"

next-i18next.config.js
```diff
@@ -0,0 +1,47 @@
...
+
+// resolve the path where translation files
+// for different languages are stored.
+// These files are used by the i18n library
+// to provide translated content for the application.
+const languageFilesPath = path.resolve("./public/locales");
+
+// define the namespaces used for
+// organizing translations in the language files.
+// Tthe "translations" namespace contains all the
```
tdilauro: Minor - typo: "Tthe ..." -> "The ..."

README.md
```diff
@@ -363,6 +370,135 @@ test("fetches search description", async () => {
...
+6. **Save changes**  
+   Verify that the translations are correct and functioning as expected in the application. Then commit the updated translation files.
+
+### Changing the app language
+
+The LanguageSelector component lets users change the language of the application to English, French, Italian, or Spanish. Note that the language code in the URL is only visible when a language other than English is selected:
+
+- **English (default)**: `https://example.com/books` (no language code `en`)
+- **French**: `https://example.com/fr/books` (language code `fr`)
+- **Italianh**: `https://example.com/it/books` (language code `it`)
```
tdilauro: Minor - typo - "Italianh" -> "Italian"

src/components/__tests__/LanguageSelector.test.tsx
```diff
@@ -0,0 +1,170 @@
...
+    locale,
+    asPath: "/test",
+    push
+  });
+
+  // return the mocked push function for assertions in tests
+  return push;
+};
+
+// define helper function to set up the test enviroment
```
tdilauro: Minor - typo


```suggestion
// define helper function to set up the test environment
```

src/components/__tests__/LanguageSelector.test.tsx
```diff
@@ -0,0 +1,170 @@
...
+const setup = (locale: string) => {
+  // fist mock the router
```
tdilauro: Minor - typo

```suggestion
  // first mock the router
```

package.json
```diff
@@ -59,16 +64,20 @@
     "dompurify": "^3.4.11",
     "downloadjs": "^1.4.7",
     "history": "^4.10.1",
+    "i18next": "^23.10.1",
```
tdilauro: Do we need to import this explicitly? It's a transitive dependency of `next-i18next` and we don't seem to use it directly. We could end up with multiple out-of-sync versions available if we accidentally import from the wrong package.

package.json
```diff
@@ -59,16 +64,20 @@
...
     "moment": "^2.30.1",
     "next": "^15.5.18",
+    "next-i18next": "^12.1.0",
+    "next-transpile-modules": "^10.0.1",
     "node-fetch": "^2.6.13",
     "opds-feed-parser": "^0.1.0",
     "react": "^18.3.1",
     "react-dom": "^18.2.0",
     "react-hook-form": "^7.53.1",
+    "react-i18next": "^12.3.1",
```
tdilauro: Same here. Do we need to import this explicitly? It's a transitive dependency of `next-i18next` and we don't seem to use it directly.

package.json
```diff
@@ -59,16 +64,20 @@
...
     "downloadjs": "^1.4.7",
     "history": "^4.10.1",
+    "i18next": "^23.10.1",
     "js-cookie": "^3.0.7",
     "jsdom": "^16.2.2",
     "lodash": "^4.18.1",
     "moment": "^2.30.1",
     "next": "^15.5.18",
+    "next-i18next": "^12.1.0",
+    "next-transpile-modules": "^10.0.1",
```
tdilauro: Do we really need `next-transpile-modules`? It was deprecated in Next 13 and I replaced its usage in `next.config.js` with the built in `transpilePackages` a few months ago when I was trying to improve start up times.

README.md
```diff
@@ -363,6 +370,135 @@ test("fetches search description", async () => {
...
+
+   This command prints a list of hardcoded strings, that are not yet wrapped in the translation function (`t`). These strings probably need to be translated, too.
+
+2. **Run the `extract` command**
+
+   ```bash
+   npm run translations:extract
+   ```
+
+   This command will scan the specified directories for translation keys used in your components and update the three `translations.json` files in the `public/locales` directory.
```
tdilauro: Will there be three files or four? Maybe we could just omit the number of files so that we don't have to keep it up to date.

README.md
```diff
@@ -363,6 +370,135 @@ test("fetches search description", async () => {
...
+
+3. **Run the `sync` command (optional)**
+
+   ```bash
+   npm run translations:sync
+   ```
+
+   This command will compare the French (`fr`), Italian (`it`), and Spanish (`es`) translation files against the English (`en`) file. It will add any missing keys from the English file to the French, Italian, and Spanish files and remove any extra keys that are not present in the English file.
```
tdilauro: Similar comment here: do we need to enumerate all of the available languages everywhere? We'll have to keep all of these sections up to date whenever we change the mix. Could we just list all of them in one place and just refer to "the translations" or "the non-English translations" elsewhere to make this a little less error prone for our future selves?

tdilauro (approved):
Looking at the list of commits here, this one must've been quite the clowder of cats to herd.

One concern -- possibly unfounded -- I have is that we might be inheriting some outdated machinery that we will eventually need to upgrade to more modern approaches. That is probably fine in the interest in near-term pragmatism of getting something working as a starting point, but it's worth getting a sense of the tech debt we're taking on and documenting that in the PR description and perhaps in a ticket for future follow up.

I'm going to approve this now with some comments/questions, since aside from a small layout issue, I don't think what's here is going to break any other functionality.

PR #180, opened 2026-08-09
Clean up async event warnings in LanguageSelector tests

tdilauro:
## Description

Silences a bunch of React `act(...)` warnings that the language selector tests printed on every test run. The tests now wait for the language selector to finish settling before making assertions, instead of checking it the instant it renders.

No test coverage or behavior changed, and no application code was touched.

## Motivation and Context

The dropdown library used by the language selector finishes some of its setup work in a microtask just after the component's first render, rather than during it. The tests were synchronous and asserting immediately, so that extra bit of work landed after the test had already moved on. That made React unhappy, causing it to warn.

Nothing was actually broken, but the warnings were pumping a lot of stack traces into the test output, making it harder to see real failures and making the logs harder to read and understand at glance.

## How Has This Been Tested?

- `npm run test:ci` — all tests pass across both suites with zero `act(...)` warnings remaining in the output.

## Checklist:

- [x] I have updated the documentation accordingly (inline).

### DIFF
--- a/src/components/__tests__/LanguageSelector.test.tsx
+++ b/src/components/__tests__/LanguageSelector.test.tsx
@@ -1,5 +1,5 @@
 import { mockUseTranslation } from "test-utils/mockUseTranslation";
-import { screen, setup as customSetup } from "test-utils";
+import { act, screen, setup as customSetup } from "test-utils";
 import LanguageSelector, { Language } from "components/LanguageSelector";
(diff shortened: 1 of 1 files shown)


PoorBillyPilgrim (approved):
Looks good @tdilauro -- thanks for cleaning up after whoever left this mess 😬

PR #181, opened 2026-08-12
Upgrade Next & React to recent versions (PP-4935)

tdilauro:
## Description

- Upgrade Next.js 15.5 to 16.2 and React 18 to 19.2, with matching `@types/react*`, `react-is`, and tooling bumps (`@next/bundle-analyzer`, `@next/eslint-plugin-next`, `@axe-core/react`).
- Stay on webpack via `--webpack` flags (Turbopack is the Next 16 default, but we have a custom webpack config); the custom HTTPS dev server opts out the same way.
- Migrate SWR 1 to 2: array keys are now passed to fetchers as a single tuple, so array-key call sites use named fetcher wrappers that destructure the key. Added a unit test covering the wrapper's argument forwarding.
- Add an npm override forcing `@nypl/design-system-react-components` onto the root React; it bundles React 16 as a hard dependency, and React 19 rejects elements created by a foreign React copy.
- Remove unused `react-redux`.
- Apply React 19 type fixes (`React.JSX` namespace, explicit `RefObject<T | null>`)
- Fix the always-false `REACT_AXE` webpack exclusion in `next.config.js`.
- Stop tracking the auto-generated `next-env.d.ts` (Next 16 rewrites it differently for dev vs. build, causing constant churn).

## Motivation and Context

Keeps the app on supported framework versions so we continue receiving security patches and bug fixes.

[Jira PP-4935]

## How Has This Been Tested?

- All tests pass locally.
- [CI checks](https://github.com/ThePalaceProject/web-patron/actions/runs/31558271081) pass.

## Checklist:


PoorBillyPilgrim (approved):
LGTM 🚀

PR #183, opened 2026-08-13
Feature flag gates language selection support (PP-4937)

tdilauro:
Adds a PALACE_CPW_FEATURE_LANGUAGE_SELECTOR environment variable feature flag (default off) that gates all patron-facing language support, following the existing PALACE_CPW_FEATURE_OPDS2 pattern:

- Language selector: with the flag off, the selector in the catalog header is not rendered at all, so it is absent from the accessibility tree.
- Automatic locale detection: with the flag off, the NEXT_LOCALE cookie and Accept-Language header are stripped from inbound requests before Next.js sees them, so every patron is served the default (English) locale. This is done dynamically at the HTTP layer in instrumentation.ts because the standalone build bakes next.config.js values in at build time, so i18n.localeDetection cannot follow a runtime environment variable. A patron's saved locale cookie is ignored, not cleared, so their choice resumes working when the flag is enabled.
- Locale-prefixed URLs: a new request proxy (src/proxy.ts, Next 16's successor to the middleware convention) 307-redirects bookmarked / hand-typed / etc. locale URLs (e.g. /es/...) to the same path in the default locale.

With the flag on, behavior is exactly as delivered in PP-4708. The i18n runtime, message catalogs, and tooling remain active in both flag states. The flag is documented in the README and .env, its state is logged at server startup, and the flag env var names moved to src/constants/env.ts so the proxy can import them without pulling in the Node-only config loader.

Motivation and Context

PP-4708 shipped the i18n framework, including a patron-facing language switcher, but other elements of the localization epic is still in progress, so the non-English catalogs are stubs that fall back to English. A patron who selects one of those languages might be confused by the lack of translation. This flag can hide the switcher and prevents any locale other than English from activating, so the framework can keep shipping to production as the localization work lands incrementally and invisibly. The flag will be removed when the epic is finished.

[Jira PP-4937]

How Has This Been Tested?

- New / updated unit tests to cover the new functionality.
- All tests and other checks pass locally.
- [CI checks](https://github.com/ThePalaceProject/web-patron/actions/runs/31748494566) pass.

Checklist:

### DIFF
--- a/src/server/__tests__/appConfig.test.ts
+++ b/src/server/__tests__/appConfig.test.ts
@@ -403,6 +403,40 @@ describe("config parsing", () => {
     });
   });
 
+  // --- enableLanguageSelector ---
+
+  describe("enableLanguageSelector", () => {
+    it("defaults to false when PALACE_CPW_FEATURE_LANGUAGE_SELECTOR is not set", async () => {
+      delete process.env.PALACE_CPW_FEATURE_LANGUAGE_SELECTOR;
+      expect((await load(MINIMAL_YAML)).enableLanguageSelector).toBe(false);
+    });
+
+    it.each([
+      ["true", true],
+      ["false", false]
+    ])(
+      "is %s when PALACE_CPW_FEATURE_LANGUAGE_SELECTOR is %s",
+      async (raw, expected) => {
+        process.env.PALACE_CPW_FEATURE_LANGUAGE_SELECTOR = raw;
+        expect((await load(MINIMAL_YAML)).enableLanguageSelector).toBe(
+          expected
+        );
+      }
+    );
+
+    it("rejects unrecognized PALACE_CPW_FEATURE_LANGUAGE_SELECTOR values", async () => {
+      process.env.PALACE_CPW_FEATURE_LANGUAGE_SELECTOR = "yes please";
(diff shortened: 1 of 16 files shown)


src/server/__tests__/appConfig.test.ts
```diff
@@ -403,6 +403,40 @@ describe("config parsing", () => {
...
+      async (raw, expected) => {
+        process.env.PALACE_CPW_FEATURE_LANGUAGE_SELECTOR = raw;
+        expect((await load(MINIMAL_YAML)).enableLanguageSelector).toBe(
+          expected
+        );
+      }
+    );
+
+    it("rejects unrecognized PALACE_CPW_FEATURE_LANGUAGE_SELECTOR values", async () => {
+      process.env.PALACE_CPW_FEATURE_LANGUAGE_SELECTOR = "yes please";
```
PoorBillyPilgrim, line 428: But it asked so nicely

PoorBillyPilgrim (approved):
👍 looks good, @tdilauro

src/server/__tests__/appConfig.test.ts
tdilauro: Gotta have some fun in this line of work! 😂

PR #184, opened 2026-08-17
Move English strings into translation files (PP-4709)

PoorBillyPilgrim:
# Description
The next step in internationalizing `web-patron` involves moving all hard-coded English strings into translation files. Translation keys have been created that map to English strings:
```json
"accountMenu.account": "Account",
"accountMenu.closeMenu": "Close account menu",
"libraryFilterList.librariesMatched_one": "{{count}} library shown, best matches first",
"libraryFilterList.librariesMatched_other": "{{count}} libraries shown, best matches first",
```

### Features & Notable Changes
- All hard-coded strings have been replaced with `t(...)` functions and `<Trans />` components provided by `next-i18next/pages`.
  - `t(...)` functions accept a translation key as their first argument. A default value can be provided as a fallback when that function's translation key doesn't exist in the current locale; otherwise the translation key itself would be rendered on the screen. This default value can be provided as standalone string or as a property of the options object, which can take the form `defaultValue` or as a variation of one of the several plural form options that exist, e.g. `defaultValue_one`, `defaultValue_two`, `defaultValue_many`. The second argument can either be the default value or the options object. If default value string is provided in the second argument, the options object can be provided as a third argument.
  - The `<Trans />` component can accept JSX elements or native HTML elements as children, which can be useful when interpolating link elements or certain semantic elements like `<b />` or `<em />`.
- OPDS parsers now collect raw values for `published` and `duration` from entries and move formatting to the components in which these values are rendered to the UI.
  - A new date utility leverages the browser-native `Intl.DateTimeFormat` API to format `published` values based on locale. A new duration utility uses a polyfill from `@formatjs/intl-durationformat` to provide a locale-aware format for `duration`.
  - > NOTE: While `Intl.DurationFormat` is a preferred, browser-native solution, it only reached baseline availability in March 2025 and is not available in Node 22. Opting for a polyfill ensures a wider browser acceptance.
- A new `common.json` translation file stores translation keys and translations for commonly used terms and phrases that do or likely will be found in 2+ components. 
- Newly added `i18n-unused` cli helps identify unused translation keys found in `translations.json` or `common.json`.
  - As it stands, the `i18next.config.ts` sets `removeUnusedKeys: false` which requires devs to manually remove any unused keys. Since `i18n-unused` is framework agnostic, a custom key checker (`tools/i18n-unused/dynamicKeyChecker.js`) was created to help parse `i18next`-specific keys, particularly in such cases where string interpolation or plural forms might cause false positives to surface.
- `i18next/no-literal-string` was enabled in `.eslintrc.js` to find any hard-coded strings not yet wrapped in `t(...)` or `<Trans />`. `"should-validate-template": true` catches interpolated strings and `"jsx-attributes"` catches component props that might be passed down to render functions, like `title` or `detail`. 
  - > NOTE: the linting rule is all-or-nothing when it comes to identifying hard-coded strings in functions. This is not an issue for most functions because they don't render to the UI, but is an issue for state setters (like `setErrorMessage`) that _do_ render to the UI. Some strings added might not be caught by the linter and devs should be aware of that limitation.

### Documentation
- Documentation provides recommended workflows for updating and adding translations during development
  - Clarifies how to use scripts and flags from `i18next-cli` and `i18next-unused` for updating the `en` translation files while updating
  - Provides guidance on consistent naming schemes for translation keys in `translations.json` and `common.json`

### Fixes
- Translation props are now present on all pages through `getServerSideProps` and `getStaticProps`
  - A previous PR (#174) erroneously assumed the i18next context and translation props were being passed through to every page. Local testing revealed that pages that received `withAppProps` and `withAppPropsSSR` only received translation props on successful cases. That is to say, translation props were not provided in the `catch` blocks of either function and thus error components could not access i18next. Similarly, pages that do not receive props from `withAppProps` and `withAppPropsSSR` did not have any translation props declared. `translationProps.ts` is a reusable utility to pass along translation props and context when `withAppProps` and `withAppPropsSSR` are not utilized.

### CI/CD
- `npm run translations:ci` has been added to the `Lint` job in `test-and-build.yml`.
  - This checks for any translation keys (in `t(...)` or `<Trans />`) that are missing from any translation file.
## Motivation and Context

This is the next step in fully internationalizing `web-patron`. The changes made here set the stage for adding all non-English translations. Since English is the default language, all translation keys and translations are keyed off of what is provided in this setup. 

[Jira PP-4709](https://ebce-lyrasis.atlassian.net/browse/PP-4709)

## How Has This Been Tested?

- Existing tests have been updated
- New `next-i18next` test utilities (`mockUseTranslation.ts` and `mockTrans.ts`) allow for reliable mocking of `useTranslation` and `<Trans />` in tests that directly leverage or rely on those utilities to function.
- Three new tests for `src/pages/_app.tsx`, `src/pages/index.tsx`, and `src/pages/[library]/[workId].tsx` were added to server-specific tests in the `tests` directory to ensure the new `getTranslationProps` helper correctly passes down the i18next context. `jest.config.node.js` was updated to transform any `.(tsx|jsx)` imports for these files.
- Tests related to the new `date.ts` and `duration.ts` utilities have been added to ensure the new locale-specific values produced by these utilities change as expected based on the current app-wide locale.

## Checklist:


src/components/bookDetails/index.tsx
```diff
@@ -145,18 +176,26 @@ export const BookDetails: React.FC = () => {
...
+    <div sx={{ my: 2 }} className={className}>
+      <H2 sx={{ mb: 2, variant: "text.headers.tertiary" }}>
+        {t("bookDetails.summary", "Summary")}
+      </H2>
+      <div
+        dangerouslySetInnerHTML={{
+          __html:
+            book.summary ??
+            t("bookDetails.summaryNotProvided", "Summary not provided.")
+        }}
```
tdilauro: This worries me because the translated string is also treated as raw HTML. Could we pull this apart so that only our `book.summary` is treated as HTML and the translation just gets treated as regular text?

src/hooks/useBorrow.ts
```diff
@@ -15,7 +17,9 @@ export default function useBorrow(isBorrow: boolean) {
   const [isLoading, setLoading] = React.useState(false);
   const { error, handleError, setErrorString, clearError } = useError();
 
-  const loadingText = isBorrow ? "Borrowing..." : "Reserving...";
+  const loadingText = isBorrow
+    ? t("actions.borrowing", "Borrowing...")
+    : t("actions.reserving", "Reserving...");
```
tdilauro: Any reason not to translate the `buttonLabel` strings on the next line? They represent important functionality.

src/components/bookDetails/index.tsx
```diff
@@ -76,14 +82,17 @@ export const BookDetails: React.FC = () => {
...
                 {book.title}
                 {book.subtitle && `: ${book.subtitle}`}
               </H1>
 
               <Text variant="text.callouts.regular">
-                by&nbsp;
-                {getAuthors(book)?.join(", ") ?? "Unknown"}
+                {t("bookDetails.by", "by")}&nbsp;
+                {getAuthors(book, t)?.join(", ") ??
+                  t("bookDetails.unknown", "Unknown")}
```
tdilauro: This block is pulling together a bunch of pieces and seems to be imposing an English language order to the phrase. It seems possible from some of the string in the existing translation to just have something akin to a template string with placeholders in the translation files, so that the ordering could be determined per-language.

I notice this somewhere else in the PR, but I can't remember where.

public/locales/en/translations.json
```diff
@@ -1,7 +1,172 @@
...
+  "reportProblem.errorMessage": "Error: {{message}}",
+  "reportProblem.problemWasReportedConfirmation": "Your problem was reported. Thank you!",
+  "reportProblem.reportProblem": "Report a problem",
+  "reportProblem.submit": "Submit",
+  "search.ariaLabel": "Enter search keyword or keywords",
+  "search.placeholder": "Enter an author, keyword, etc...",
+  "search.search": "Search",
+  "signedOut.returnToCatalog": "Return to Catalog",
+  "signedOut.securityNotice.body1": "You have been signed out of this application, but your session with the authentication provider may still be active.",
+  "signedOut.securityNotice.body2": "This could lead to you (or someone else) being signed back into your account without needing to provided your credentials; or it could prevent you from signing into a different account.",
```
tdilauro: typo: "provided" -> "provide"

I checked the current `/<slug>/signed-out` page and it's there, too. 🤦🏽‍♂️ My bad, but we should probably fix it.

README.md
```diff
@@ -410,52 +418,113 @@ Key settings:
...
 To extract translation keys from your component source code and to update the `translations.json` files, follow these steps:
 
+0. **Run `npm run lint`**
+
+   ```bash
+   npm run lint
+   ```
+
+   This confirms that any new keys are namespaced correctly for the file they live in. Do this before extracting, so that a mis-prefixed key never reaches the `translations.json` files.
+
```
tdilauro: I'm confused by the two lint commands (new item 0 and existing item 1). Would you clarify these? the first one appears to be a regular code linter, but it has a translation-related description.

next.config.js
```diff
@@ -69,10 +69,22 @@ const config = {
...
+    // can't) depend on either of these in client-side code.
+    //
+    // fs:     server-only modules that read from disk.
+    //
+    // module: pages/_error.tsx imports dataflow/translationProps, which pulls in
+    //         next-i18next's serverSideTranslations, which calls createRequire
+    //         from "module" to load next-i18next.config.js. _error may only use
+    //         getInitialProps, and Next does not strip that from the client
+    //         bundle the way it strips getStaticProps, so even though the IS_SERVER
+    //         guard in _error.getInitialProps means it never runs there.
```
tdilauro: I'm having trouble parsing this sentence. That said, I think I understand roughly what it means to say. Might be simpler to just mention that we need this to prevent build problems for the client/browser code when using next-i18next, so it's clear that it shouldn't be removed.

src/auth/Login.tsx
```diff
@@ -32,27 +33,33 @@ export default function Login(): React.ReactElement {
...
-            </ExternalLink>
-            .
-          </Text>
-        )}
+        <Trans i18nKey="login.noAuth">
+          This Library does not have any authentication configured.{" "}
+          {helpEmail && (
+            <Text>
+              If this is an error, please contact your site administrator via
+              email at:{" "}
```
tdilauro: Is this going to render what we expect if a `helpEmail` is present? It doesn't look like that case gets tested.

src/components/SignOut.tsx
```diff
@@ -127,22 +129,27 @@ export const SignOut: React.FC<SignOutProps> = ({
...
+          )}
+        </p>
         <Stack sx={{ justifyContent: "center" }}>
           <Button variant="ghost" color="ui.gray.dark" onClick={dialog.hide}>
-            Cancel
+            {t("actions.cancel", "Cancel", { ns: "common" })}
           </Button>
           <Button color="ui.error" onClick={signOutAndClose}>
-            Confirm Sign Out
+            {t("signOut.signOut", "Sign Out")}
```
tdilauro: This was "Confirm Sign Out" before. This has the same key and default text as the "initiate sign out" button. Was that intentional?

public/locales/en/translations.json
```diff
@@ -1,7 +1,172 @@
...
+  "utils.book.mediumBook": "Book",
+  "utils.book.mediumEbook": "eBook",
+  "utils.book.more_one": "{{count}} more",
+  "utils.book.more_other": "{{count}} more",
+  "utils.book.onHold": "You have this book on hold.",
+  "utils.book.onHoldUntil": "You have this book on hold until {{until}}.",
+  "utils.book.onLoanUntil": "You have this book on loan until {{availableUntil}}.",
+  "utils.book.unknownAuthor": "Unknown Author",
+  "utils.book.unknownAuthors": "Authors unknown",
+  "utils.books.positionInQueue": "{{position}} patrons ahead of you in the queue.",
```
tdilauro: Should this one be "utils.book..." like the others?


```suggestion
  "utils.book.positionInQueue": "{{position}} patrons ahead of you in the queue.",
```

tdilauro (commented):
Impressive amount of work in here and this is in pretty good shape, especially considering how big it is. 👏🏽 

I left a few comments / questions to consider below.

src/hooks/useBorrow.ts
PoorBillyPilgrim: No reason -- just missed it due to the way the linter doesn't pick up strings as variables.I caught this earlier when I was working on the ticket for the non-English translations. I'll update it in this PR.

src/components/SignOut.tsx
PoorBillyPilgrim: Yes, this was intentional. When going through this component, I felt that "Sign Out" conveyed the same message as "Confirm Sign Out" and opted for consistent verbiage for the "sign out" user journey

src/auth/Login.tsx
PoorBillyPilgrim: I'll add tests in `Login.test.tsx` to check for what we expected to be rendered

README.md
PoorBillyPilgrim: The two commands are complementary. I'll add their differences to `README.md`

tdilauro (approved):
Looks good! Nice work! 🏁

PR #185, opened 2026-08-20
Make i18n config visible to containerized app (PP-4961)

tdilauro:
## Description

Pass the `next-i18next` config to `serverSideTranslations` explicitly instead of relying on it being found on disk at request time.

- Every page builds its i18n props through one helper, so this is a single call site.
- Adds a type declaration so the CommonJS config can be imported under `allowJs: false`.
- Also copies the config into the runtime image in case a future call site omits it.

## Motivation and Context

Every server-rendered page returned an error page in the dev deployment, and locally when running the standalone server directly.

Left to itself, next-i18next resolves `./next-i18next.config.js` against the working directory on every render. Nothing in our code referenced the file, so it only ever reached the standalone bundle because Next's file tracer inferred it. Up until Next 16.3.0 replaced it (`@vercel/nft` 0.27.1 -> 1.10.2). Now it no longer emits "working-directory-relative" assets discovered in dependency code, and the file stopped being copied. Importing the config removes the dependency on both the tracer and the working directory.

[Jira PP-4961]

## How Has This Been Tested?

- Manual testing in local dev environment. Specifically:
  - Production build run through the standalone server with the config file absent: the multi-library home page, library pages, and the `/en` locale route all render with translations and no errors.
  - `npm run dev` verified for the same routes, with the language selector flag both on and off.
- All tests / checks pass locally.
- [CI tests / checks](https://github.com/ThePalaceProject/web-patron/actions/runs/32385569818) pass.

## Checklist:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,5 +28,8 @@ ENV NODE_ENV=production \
 COPY --from=builder /app/_next/standalone ./
 COPY --from=builder /app/_next/static ./_next/static
 COPY --from=builder /app/public ./public
+# next-i18next resolves ./next-i18next.config.js against the working directory
+# when a call site does not pass the config explicitly.
+COPY --from=builder /app/next-i18next.config.js ./
 EXPOSE 3000
 CMD ["node", "server.js"]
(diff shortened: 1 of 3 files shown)


PoorBillyPilgrim (approved):
Looks good, thanks @tdilauro

PR #186, opened 2026-08-21
Memoize getLibraries merge (PP-5018)

tdilauro:
## Description

Memoize the merged library list in `getLibraries`. The merge result is reused until a registry cache entry is written, instead of allocating a fresh object spanning every library on each call. A version counter on the shared registry state invalidates the memo on every cache write (including across Next.js bundle contexts), and the memo is also keyed on the config object reference. Because callers now share one object, the return type is `Readonly` to enforce the no-mutation contract. Memoization is per-process; instances behind a load balancer each keep their own cache and memo.

## Motivation and Context

`getLibraries` runs on every page render and every `/api/libraries` request, but its result only changes when a registry refresh writes new data (at most once per refresh interval). With a large registry, rebuilding the merged map per request created constant allocation and GC pressure on the server. This addresses one of the findings from the server-side memory investigation (PP-4959).

[Jira PP-5018]

## How Has This Been Tested?

- New tests verify:
  - that repeated calls return the same object while the cache is unchanged;
  - that the merge is recomputed after a refresh writes new data;
  -  and that a memoized result is not reused for a different config object. 
- All tests / check pass locally and [in CI](https://github.com/ThePalaceProject/web-patron/actions/runs/32505496838). 

## Checklist:

### DIFF
--- a/src/server/__tests__/libraryRegistry.test.ts
+++ b/src/server/__tests__/libraryRegistry.test.ts
@@ -927,6 +927,87 @@ describe("getLibraries", () => {
     await getLibraries(config);
     expect(fetchMock).toHaveBeenCalledTimes(1);
   });
+
+  it("returns the same merged object across calls while the cache is unchanged", async () => {
+    const feed = makePagedFeed([
+      { id: "urn:uuid:a", title: "A", authDocUrl: "https://a.example.com/" }
+    ]);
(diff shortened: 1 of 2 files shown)


PoorBillyPilgrim (approved):
LGTM

PR #188, opened 2026-08-23
Reduce prefetching on multi-library pages (PP-5018)

tdilauro:
## Description

- Turn off `next/link`'s viewport prefetch for the library links on the multi-library home page, so a library is prefetched only when a patron shows intent by hovering it or by tabbing to it.
- Add a debounced prefetch when a link receives keyboard focus. Next.js prefetches on hover and touch but not on focus, so this provides comparable behavior for keyboard patrons. The debounce keeps links that are tabbed through quickly enough from being prefetched.

**Note**: Next.js disables prefetching outside a production build, so the change is only observable in a production build.

## Motivation and Context

The multi-library home page lists every configured library as a link, and Next.js prefetched every one that scrolled into view. Library pages are generated on demand, and each render fetches that library's authentication document from its circulation manager. Opening the page therefore triggered one on-demand render and one upstream fetch per visible library, which in a large window can be quite expensive. The work is wasted, since a patron goes to one library. This also created a lot of logging output.

[Jira PP-5018]

## How Has This Been Tested?

- Manual testing with build and standalone server.js in local dev environment.
- New tests for the added functionality.
- Tests and checks pass locally and [in CI](https://github.com/ThePalaceProject/web-patron/actions/runs/32612884764).

## Checklist:

### DIFF
--- a/src/components/LibraryHomeLink.tsx
+++ b/src/components/LibraryHomeLink.tsx
@@ -1,6 +1,13 @@
 /* eslint-disable jsx-a11y/anchor-is-valid */
 import * as React from "react";
 import Link from "next/link";
+import { useRouter } from "next/router";
+
+/*
+ * Long enough to wait out an auto-repeating Tab key, yet short enough to
+ * do prefetch before a patron presses Enter after settling on a library.
+ */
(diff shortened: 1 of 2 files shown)


jonathangreen (approved):
Looks great to me! 🚀

PR #189, opened 2026-08-23
Config clean up and linting additions

tdilauro:
## Description

- Extend the `lint` and `lint:fix` scripts to cover files at the repository root. The previous glob required a directory segment, so root config files were never linted.
- Fix the lint errors this uncovered: an empty `catch` block and an unused webpack parameter in the Next.js config, and a `var` declaration in the HTTPS dev server.
- Remove a dead `generateBuildId` from the Next.js config. A second definition later in the same file overrode it, so it never ran. Build IDs are unaffected.

No runtime behavior changes.

## Motivation and Context

Root-level config files were silently exempt from linting, so errors there went unnoticed and edits to those files got no feedback. Covering these makes the existing rules apply everywhere and keeps the config files from drifting again.

The extra, dead `generateBuildId` was found while fixing those errors.

## How Has This Been Tested?

- `npm run lint` passes.
- Confirmed the widened glob actually covers root files by temporarily introducing an error in the Next.js config and seeing lint catch it.
- Confirmed the Next.js config still loads and exports the same keys.
- `npm run type-check` and the full Jest suite pass.

## Checklist:


PoorBillyPilgrim (approved):
👍

